Issue #3, opened 2019-01-04
Document port binds and what they are used for


sparky8251:
I know it at least binds 8096/tcp (HTTP) and 8920/tcp (HTTPS), but we should check if it binds anything else and document those as well.

sparky8251:
Ah! So they could very well be listening but there's no way of telling without digging into the code. All the more reason to investigate and find out.

sparky8251:
7359/udp seems to be part of the DLNA/Chromecast implementation.

If the server implementations, you can see it bind that udp port and this related code:

```
AddMessageResponder("who is EmbyServer?", true, RespondToV2Message);
AddMessageResponder("who is MediaBrowserServer_v2?", false, RespondToV2Message)
```

sparky8251:
[And going by this bit of code](https://github.com/jellyfin/jellyfin/blob/f8a720d3d8adbdb1f092a42e592dae37ba3f25bb/Emby.Server.Implementations/LiveTv/TunerHosts/HdHomerun/HdHomerunUdpStream.cs#L55), the random udp port is for at least a specific bit of LiveTV support (possibly all).

sparky8251:
Looks like the random port is what a network tuner is told to send the LiveTV streams to.

JustAMan:
Closed by #4.

PR #4, opened 2019-01-04
Add admin docs on port mappings

sparky8251:
Described what each is used for and if they are admin configurable or not.

Fixes issue #3


joshuaboniface:
Good idea starting admin docs, thouth I'm not sure where we want to draw the line between "user" and "administrator".

JustAMan:
Easy-peasy, `user` is only watching/listening stuff, not configuring anything except his/her own password, picture and probably theme.

PR #5, opened 2019-01-05
Added equivs to deb package prerequisits

ploughpuff:
In a clean install of Ubunt 18.10 when trying to build jellyfin from source got the following error:

ploughpuff@vb-jelly:~/jellyfin$ sudo mk-build-deps -i
mk-build-deps: You must have equivs installed to use this program.


ploughpuff:
My proposed changes here are insufficient to allow Ubuntu to build successfully.

joshuaboniface:
Yup that's definitely a bug, probably something already installed on my systems. That's a good change though.

joshuaboniface:
@ploughpuff Actually if you can let me know what does work, I want to tweak that section a bit and split out "install dependencies" from the "install the Microsoft repo" part.

PR #23, opened 2019-01-21
Update colors

nvllsvm:
image


JustAMan:
Now these look nice, can we fill the missing branding pieces?https://jellyfin.readthedocs.io/en/latest/developer-docs/branding-guide/#jellyfin-colours

PR #34, opened 2019-02-07
Add donation badge and reorganize badges

joshuaboniface:
Add the OpenCollective badge to the main page and :shipit:.


JustAMan:
Maybe we should wait with merge till same PR lands in main repo https://github.com/jellyfin/jellyfin/pull/834 ?

PR #40, opened 2019-02-16
Add example reverse proxy configs

mcarlton00:
Going for the minimal required to get each service up and running.  Each service was tested by authenticating and playing videos through the web UI and android app.

### DIFF
--- a/docs/administrator-docs/reverse-proxy.md
+++ b/docs/administrator-docs/reverse-proxy.md
@@ -33,6 +33,7 @@ Three popular options for reverse proxy systems are [Apache](https://httpd.apach
     SSLEngine on
     SSLCertificateFile /etc/letsencrypt/jellyfin.example.com/fullchain.pem
     SSLCertificateKeyFile /etc/letsencrypt/jellyfin.example.com/privkey.pem
+    Protocols h2 http/1.1
 
     ErrorLog /var/log/apache2/jellyfin.example.com-error.log
     CustomLog /var/log/apache2/jellyfin.example.com-access.log combined
@@ -72,7 +73,7 @@ server {
 }
 
 server {
-    listen 443 ssl;
+    listen 443 ssl http2;
     server_name jellyfin.example.com;
     ssl_certificate /etc/letsencrypt/live/jellyfin.example.com/fullchain.pem;
     ssl_certificate_key /etc/letsencrypt/live/jellyfin.example.com/privkey.pem;
--- a/docs/administrator-docs/reverse-proxy.md
+++ b/docs/administrator-docs/reverse-proxy.md
@@ -53,6 +53,8 @@ $ sudo a2enmod ssl
 frontend jellyfin_proxy
     bind *:80
 # Note that haproxy requires you to concatenate the certificate and key into a single file
+## Note that if you're using haproxy 1.8+, you can enable http2 by swapping these lines
+##   bind *:443 ssl crt /etc/letsencrypt/live/jellyfin.example.com/complete.pem alpn h2,http/1.1
     bind *:443 ssl crt /etc/letsencrypt/live/jellyfin.example.com/complete.pem
     redirect scheme https if !{ ssl_fc }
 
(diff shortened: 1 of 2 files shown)


LeoVerto:
Adding `http2` to the nginx listen argument wouldn't hurt, would it?

mcarlton00:
I tested http2 with all 3.  Apache and Nginx work fine, but when it's enabled in Haproxy the android app can't connect.  Probably a configuration error on my part, but I can't figure it out at the moment

joshuaboniface:
Looks good, but might be worthwhile adding a bit about LetsEncrypt/Certbot as well, since a number of people have asked about it.

 I have an example for NGiNX here: https://gist.github.com/joshuaboniface/e843ed141bb69272f2e39a30961d26c7

HAProxy can also do something similar:

```
frontend http
    bind :::80 v4v6
    mode http
    option httplog
    acl url_letsencrypt path_beg /.well-known/acme-challenge/
    use_backend debX_letsencrypt if url_letsencrypt
    redirect scheme https if !url_letsencrypt !{ ssl_fc }
```

joshuaboniface:
@mcarlton00 Can you let me know how you enabled http2 in haproxy? I can test it out a bit too.

mcarlton00:
http2 on haproxy:
I just changed changed:
```
bind *:443 ssl crt /etc/letsencrypt/live/jellyfin.example.com/complete.pem
```
to
```
bind *:443 ssl crt /etc/letsencrypt/live/jellyfin.example.com/complete.pem alpn h2,http/1.1
```
which has worked on other projects I've done.

-------------------------

Thinking about it, I can add SNI into the haproxy config so it's more similar to the apache and nginx ones.
```
acl jellyfin_server hdr(host) -i jellyfin.example.com
use_backend jellyfin if jellyfin_server
```
If we include LetsEncrypt, that could be `... if jellyfin_server !url_letsencrypt`

How detailed do we want to get with the LetsEncrypt stuff?  There's just so much variation between certbot or other clients, auth methods, ports, etc.

joshuaboniface:
Thanks @mcarlton00 I'll test that out today. SNI is a good thing too, simpler configs!

For LE, yea that was just a dumb copy-paste from my live config where I send the requests to `debX` just ignore that :-P

I think just a basic "here's how you pass it through, for more advice see Certbot/LetsEncrypt docs" would be sufficient - I'm more interested in having just the basics with Certbot in there as an "easy-to-use" quickstart, and let more advanced users branch out from there.

joshuaboniface:
Actually just did a quick test with `alpn h2,http/1.1` on my HAProxy 1.8.14-1 instance, and it seems to work fine even from Android - what wasn't working on that?

mcarlton00:
I think a simple webroot auth is good for most people, but does that work with haproxy?  Since it's not a "proper" webserver, I'm not sure if it'll serve those files properly.  Would we have to specify certbot on a different port for LetsEncrypt to work?

 Just did a quick test, http2 seems to be only functional on haproxy 1.8 and newer.  So we can include it, but should definitely make a note.

joshuaboniface:
>I think a simple webroot auth is good for most people, but does that work with haproxy? Since it's not a "proper" webserver, I'm not sure if it'll serve those files properly. Would we have to specify certbot on a different port for LetsEncrypt to work?

I don't think so, I always just run Certbot in HTTP challenge mode on that target port. You would - the port you specify in Certbot must be the proxy target port from HAProxy/NGiNX.

>Just did a quick test, http2 seems to be only functional on haproxy 1.8 and newer. So we can include it, but should definitely make a note.

Oh yes it's 1.8+ only - I think 1.8 is in a lot of distros or will be soon, so it's worth mentioning explicitly.

JustAMan:
@joshuaboniface if you verified this stuff works for HAproxy, can you also approve? I can only verify `nginx` part, so I don't feel comfortable merging other stuff...

mcarlton00:
As far as just the base proxy stuff is concerned, this should be good to go now.  I probably won't have time to get back to this until the weekend.  Do we want to merge so we can get the info out there and I'll do another PR adding LetsEncrypt stuff once I get some free time again?

PR #41, opened 2019-02-19
Add docker quickstart section

mcarlton00:
Just adding simple quick start instructions to the docker section


joshuaboniface:
Should add a bit about the cache dir volume too (see jellyfin/jellyfin#920), but otherwise LGTM!

PR #44, opened 2019-03-01
Add link to kodi addon install guide

mcarlton00:
Adding a link to the install guide on the jellyfin-kodi repo.  Partner PR to the improvements to the install guide [here](https://github.com/jellyfin/jellyfin-kodi/pull/11)


joshuaboniface:
I'd much rather the main doc be here and link in the repo to be honest. Is that OK?

mcarlton00:
Personally, I think that's better in terms of user friendliness.  I thought this was how we had decided to handle it, but looking back at the matrix conversation it looks like I interpreted it backwards.  I'll swap it around later today

PR #49, opened 2019-03-08
Revised Docker instructions and installation link

cfunseth:
I've updated the homepage link to direct towards the installation instructions (rather than the Docker Hub page) as this caused some confusion on my part when initially looking for help.  This is now in line with how the other platform instructions are linked.

I've also updated the Docker instructions to use the proper mount points declared in the Dockerfile, as well as changed the media mount to a read-only bind.  

Finally, I've added a docker-compose example of deploying this container.

### DIFF
--- a/docs/administrator-docs/installing.md
+++ b/docs/administrator-docs/installing.md
@@ -20,11 +20,23 @@ The Jellyfin Docker image is available on [Docker Hub](https://hub.docker.com/r/
     `--name jellyfin \`  
     `--volume /mnt/jellyfin_config:/config \`  
     `--volume /mnt/jellyfin_cache:/cache \`  
-    `--volume /path/to/media:/mnt/library \`  
-    `--publish 8096:8096 \`  
+    `--volume /path/to/media:/media:ro \`  
     `--net=host \`  
-    `jellyfin/jellyfin:latest`
-
+    `jellyfin/jellyfin:latest`  
+  
+Alternative docker-compose example:  
+```
+    version: "3"  
+    services:  
+        jellyfin:  
+          container_name: jellyfin  
+          image: jellyfin/jellyfin:latest  
+          network_mode: "host"  
+          volumes:  
+            - /mnt/jellyfin_config:/config  
+            - /mnt/jellyfin_cache:/cache  
+            - /path/to/media:/media:ro  
+```
 ### Unraid Docker
 
 An Unraid Docker template is available in the repository.
--- a/docs/administrator-docs/installing.md
+++ b/docs/administrator-docs/installing.md
@@ -17,7 +17,6 @@ The Jellyfin Docker image is available on [Docker Hub](https://hub.docker.com/r/
     `mkdir /path/to/cache`
 3. Start the server:  
     `docker run -d \`  
-    `--name jellyfin \`  
     `--volume /path/to/config:/config \`  
     `--volume /path/to/cache:/cache \`  
     `--volume /path/to/media:/media \`  
@@ -29,7 +28,6 @@ Alternative docker-compose example:
     version: "3"  
     services:  
         jellyfin:  
-          container_name: jellyfin  
           image: jellyfin/jellyfin  
           network_mode: "host"  
           volumes:  
--- a/docs/administrator-docs/installing.md
+++ b/docs/administrator-docs/installing.md
@@ -13,16 +13,16 @@ The Jellyfin Docker image is available on [Docker Hub](https://hub.docker.com/r/
 1. Get the latest image:  
     `docker pull jellyfin/jellyfin`
 2. Create directories on the host for persistent data storage:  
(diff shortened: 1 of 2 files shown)


docs/administrator-docs/installing.md
```diff
@@ -20,11 +20,23 @@ The Jellyfin Docker image is available on [Docker Hub](https://hub.docker.com/r/
     `--name jellyfin \`  
     `--volume /mnt/jellyfin_config:/config \`  
     `--volume /mnt/jellyfin_cache:/cache \`  
-    `--volume /path/to/media:/mnt/library \`  
-    `--publish 8096:8096 \`  
+    `--volume /path/to/media:/media:ro \`  
     `--net=host \`  
-    `jellyfin/jellyfin:latest`
-
+    `jellyfin/jellyfin:latest`
```
nvllsvm: This was here before, but `:latest` is redundant - it's used when no tag is specified.

docs/administrator-docs/installing.md
```diff
@@ -20,11 +20,23 @@ The Jellyfin Docker image is available on [Docker Hub](https://hub.docker.com/r/
     `--name jellyfin \`  
     `--volume /mnt/jellyfin_config:/config \`  
     `--volume /mnt/jellyfin_cache:/cache \`  
-    `--volume /path/to/media:/mnt/library \`  
-    `--publish 8096:8096 \`  
+    `--volume /path/to/media:/media:ro \`
```
nvllsvm: I'm a fan of using `:ro` myself, but I think it technically breaks some features. Not sure how I feel about advertising it as the default.

Also - keep the example paths consistent. Either make all `/mnt/whatever` or `/path/to/whatever`.

docs/administrator-docs/installing.md
```diff
@@ -20,11 +20,23 @@ The Jellyfin Docker image is available on [Docker Hub](https://hub.docker.com/r/
...
-
+    `jellyfin/jellyfin:latest`  
+  
+Alternative docker-compose example:  
+```
+    version: "3"  
+    services:  
+        jellyfin:  
+          container_name: jellyfin  
+          image: jellyfin/jellyfin:latest
```
nvllsvm: `:latest` is redundant - it's used when no tag is specified.

docs/administrator-docs/installing.md
```diff
@@ -20,11 +20,23 @@ The Jellyfin Docker image is available on [Docker Hub](https://hub.docker.com/r/
...
-    `jellyfin/jellyfin:latest`
-
+    `jellyfin/jellyfin:latest`  
+  
+Alternative docker-compose example:  
+```
+    version: "3"  
+    services:  
+        jellyfin:  
+          container_name: jellyfin
```
nvllsvm: This is unnecessary for Docker Compose.

Ex. You can still do this without `container_name`.

```
docker-compose exec jellyfin ls
```

docs/administrator-docs/installing.md
```diff
@@ -20,11 +20,23 @@ The Jellyfin Docker image is available on [Docker Hub](https://hub.docker.com/r/
...
+Alternative docker-compose example:  
+```
+    version: "3"  
+    services:  
+        jellyfin:  
+          container_name: jellyfin  
+          image: jellyfin/jellyfin:latest  
+          network_mode: "host"  
+          volumes:  
+            - /mnt/jellyfin_config:/config
```
nvllsvm: Make the example paths consistent.

docs/administrator-docs/installing.md
cfunseth: I'll remove the `latest` tag. Do you see any value in documenting the different tag options here or just leave it simple and let the user reference the Docker Hub page?

docs/administrator-docs/installing.md
cfunseth: I had changed this due to the recent Emby scare, but you're correct that this conflicts with Jellyfin's ability to delete media from the menu options.  I'll remove the `:ro` tag.

The paths used here were from the original example and follow the existing installation steps.  I will modify them to read as `/path/to/config` and `/path/to/cache` and revise the instructions to be a bit more clear.

docs/administrator-docs/installing.md
cfunseth: I agree that this isn't necessary, but the resulting container will have a name of `<stack_name>_jellyfin_1` that I don't feel is as intuitive.

Also, the original example used the `--name` parameter, so specifying the `container_name` keeps the two methods consistent with one another.

nvllsvm: There's little reason to keep them consistent since the docker-compose command negates the need to have the container name be friendly. Please remove.

docs/administrator-docs/installing.md
nvllsvm: No need to document tag options here.

Issue #45, opened 2019-03-01
LetsEncrypt examples in reverse proxy page


mcarlton00:
As discussed in #40, improve the reverse proxy configs to have examples of how to use LetsEncrypt authentication.

Issue #48, opened 2019-03-06
Cannot setup a reverse proxy using the documentation


pyluyten:
I used this documentation : https://jellyfin.readthedocs.io/en/latest/administrator-docs/reverse-proxy/ .

Some items were not clear to me.
* the doc should state prerequisite : a running http server + own a domain name. If a subdomain is the only way to reverse proxy then it should also state so.
* the doc seems to assume the machine running jellyfin is also the machine running the reverse proxy. So this should be written clearly. 
* the doc should highlight which part of code snippet should be adapted, eg "jellyfin.example.com" is to be adapted.

Anyway so i figured i can add a VirualHost on my machine running jellyfin, pointed a DNS subdomain to this machine IP, added the code snippet, activated the site.

If I go to jellyfin.my-site.xxx, i will obtain : "INTERNAL SERVER ERROR'. Server log says, "[Tue Mar 05 23:28:28.924719 2019] [proxy:warn] [pid 24036] [client xx.xxx.xx.xxx:56940] AH01144: No protocol handler was valid for the URL /embywebsocket. If you are using a DSO version of mod_proxy, make sure the proxy submodules are included in the configuration using LoadModule."

At the moment i did not figured issue yet, but almost certain the doc does not explain what to do. (Missing prerequisite?)

However rather than my little issue the reason i open this ticket is to discuss the documentation.

JustAMan:
cc @mcarlton00 @joshuaboniface

mcarlton00:
Admittedly, Apache is the one that I'm the least familiar with.  I copied the config that @sauerj provided and tested it locally to make sure it works.  It sounds like you possibly didn't enable the extra modules from the section that's under the main code block for Apache.  Try running `a2enmod proxy proxy_http ssl` and then restart Apache.  Sometimes these are included by default, sometimes they need to be a manual step.  I think it depends on which distro you're using.

As for improvements:
> * running http server + own domain name

The docs are outlining how to set up a http server, so I'm not sure that should necessarily be a requirement.  I hesitate to include installation instructions just because of how much variance there is between distros and package managers.  We probably should make a note about the domain name, though.  It's really only strictly necessary if people are accessing the server from outside their local network, but it's almost certainly the most common use case we'll see.

> * machine running jellyfin is also the machine running the reverse proxy
> * highlight which part of code snippet should be adapted

[...]

PR #51, opened 2019-03-10
Reverse Proxy improvements

mcarlton00:
Added some notes about prerequisites and changed the server blocks to have a more generic format (`IP_ADDRESS` instead of `127.0.0.1`, same with server names).  Mostly following through on suggestions from #48.

Also added notes about LetsEncrypt and Certbot for each of the proxy systems that should close https://github.com/jellyfin/jellyfin-docs/issues/45.  Could use some discussion here, based on different people's opinions.  I've tested all three and was able to get new certificates, but they could probably get cleaned up further.
* I'm suggesting using the Apache and Nginx plugins to certbot instead of the standalone method [mentioned before](https://github.com/jellyfin/jellyfin-docs/pull/40#issuecomment-464976746). Both because it's a little easier on the end user, and because trying to proxypass from Apache to the certbot standalone port was frustrating me.  Still using the standalone method for haproxy, but that one works smoothly.
* Explicitly stating things that won't work without an ssl cert, as opposed to giving the full code blocks that won't let the service start right off the bat (Apache and Nginx won't start with a SSL server block with no certificates available).
* Could use opinions on the server blocks themselves.  At this moment, the example server block is enough to get a SSL certificate, but not reach JF immediately over plain http.  But if we add support for http in, the blocks are going to get a good bit longer and require more explanation about what to change after a certificate has come into play.

I'm not particularly thrilled about giving out server blocks that can't actually reach JF right away, but i'm also not thrilled about longer and more complicated instructions of how to switch.  Formatting suggestions are welcome.

One idea is to structure the page like:
```
http:
  apache:
  haproxy:
  nginx:
letsencrypt:
  apache:
  haproxy:
  nginx:
https:
  apache:
  haproxy:
  nginx:
```

### DIFF
--- a/docs/administrator-docs/reverse-proxy.md
+++ b/docs/administrator-docs/reverse-proxy.md
@@ -4,47 +4,53 @@ It's possible to run Jellyfin behind another server acting as a reverse proxy.  
 
 Three popular options for reverse proxy systems are [Apache](https://httpd.apache.org/), [Haproxy](https://www.haproxy.com/), and [Nginx](https://www.nginx.com/).
 
+**Important:** In order for a reverse proxy to have the maximum benefit, you should have a publically routable IP address and a domain with DNS set up correctly.  These examples assume you want to run Jellyfin under a sub-domain (ie: jellyfin.example.com), but are easily adapted for the root domain if desired.
+
+When following this guide, be sure to replace the following variables with your information:
+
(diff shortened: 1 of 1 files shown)


joshuaboniface:
i think this is OK. There's only so much hand-holding we can do with this, as these are big topics to cover.

PR #54, opened 2019-03-19
Script distro name in a oneliner

rkarlsba:


joshuaboniface:
Does this work consistently for every release though?

rkarlsba:
So long that lsb_release is installed, yes, and on debian and relatives (such as ubuntu), it's in there by default.

joshuaboniface:
Confirmed, this actually works properly. Time to change some other scripts that were having to parse `/etc/os-release`!

PR #60, opened 2019-03-28
Step-by-step guide to the configuration of the Jellyfin TVHeadend plugin and TVheadend

97carmine:
I have created a step-by-step guide to configure the TVHeadend plugin and Tvheadend.


dkanada:
Some screenshots might be a good idea to replace the configuration options, but this looks really thorough otherwise!

If you want to work on the docs a bit more we have a [ton of pages](https://github.com/jellyfin/jellyfin-web/search?q=github.com%2FMediaBrowser&unscoped_q=github.com%2FMediaBrowser) that still need entries in the documentation. I have been [working through them](https://github.com/jellyfin/jellyfin-docs/pull/61) but there are still a lot of gaps that need to be filled.

97carmine:
It was the initial idea, but I do not know how to add screenshots, I decided to use the configuration options.

I believe that the configuration options should not be replaced by the screenshots but that the screenshots and the configuration options complement each other.

PR #62, opened 2019-04-05
update windows build instructions

jftuga:
Fixed instructions to build `Windows on Windows`. I also added `-verbose` so that all of the `Write-Verbose` messages will be seen.


jftuga:
Can this also be pushed to:

https://jellyfin.readthedocs.io/en/latest/administrator-docs/building/#windows-on-windows

Thanks.

PR #65, opened 2019-04-09
Update installing.md

Maneeds:
Hi not sure if this is right, just signed up for GitHub after using Jellyfin for a few weeks. Found the Debian repository instructions tricky to follow as things were marked as code which I don’t think need to be for legibility to non-tech users (like me). Thought I’d try and start to give back something to the project as I can’t code. 

Hope this is ok.

### DIFF
--- a/docs/administrator-docs/installing.md
+++ b/docs/administrator-docs/installing.md
@@ -88,9 +88,9 @@ CentOS/RHEL 7 builds in RPM package format are available [here](https://repo.jel
 
 ### Repository
 
-The Jellyfin team provides a Debian repository for installation on Debian Stretch/Buster, for both `amd64` and `armhf`.
+The Jellyfin team provides a Debian repository for installation on Debian Stretch/Buster, for both amd64 and armhf.
 
-**NOTE:** Microsoft does not provide a .NET for 32-bit x86 Linux systems, and hence Jellyfin is not supported on the `i386` architecture.
+**NOTE:** Microsoft does not provide a .NET for 32-bit x86 Linux systems, and hence Jellyfin is not supported on the i386 architecture.
 
 1. Install HTTPS transport for APT if you haven't already:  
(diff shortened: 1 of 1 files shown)


joshuaboniface:
Hey there, I chose the fixed-width (i.e. "code") font for certain elements to indicate that they are literal values you may see elsewhere in code blocks, including paths and such. So for instance I say `amd64` to refer specifically to that string which is what is valid in a `sources.list` file, rather than AMD64 as I would in a normal paragraph. This is a fairly common style I see elsewhere and ensures clarity between the explanatory paragraph and the actual code blocks (which are always on a new line following a `:` character).

I'd prefer to keep this format as the standard unless a better one is suggested.

Maneeds:
@joshuaboniface  thanks for clarifying. Makes complete logical sense why you do things that way.

I do think your superior knowledge clouds the fact that us less technically inclined users can be easily confused. 

Perhaps other devs would understand why amd64 was formatted in code blocks - and now you've explained it to me, I understand too (the colon character before actual code too, makes total sense). But devs aren't the ones who will be following them docs or need that pointer.  

Ashamedly I've followed other Jellyfin instructions and pasted text formatted as code into Terminal when in fact the author was only aiming to highlight it (I don't think it helps I view them on mobile so the formatting can look a little broken). I persevered after my mistake and got things installed eventually - but would all users do that? Should instructions be foolproof to follow? Remembering the audience level I would aim for plain English steps as much as possible. 

Kodi for example does not highlight file paths but even I know that's what they are:
https://kodi.wiki/view/Userdata

(I actually think highlighting that they're file paths is fine, underline/italics,bold, but using the code block formatting as Jellyfin is doing can be confusing to the layman).

Regardless of the direction, thanks for the tips. I should be safe from further embarrassing mistakes going forward. :)

joshuaboniface:
Glad to help clarify :) I'm going to close this one out!

PR #66, opened 2019-04-09
Add instructions to run jellyfin from source

DrPandemic:
I'm not used to dotnet core so it took me some time to understand how to run it after building it from source.

Adds an instruction to the list to run it.

### DIFF
--- a/docs/administrator-docs/building.md
+++ b/docs/administrator-docs/building.md
@@ -31,6 +31,8 @@ All package builds begin with the first two steps:
     `./build --list-platforms`
     `./build <platform> all`
 
+4. The resulting archives can be found at `../bin/<platform>`
+
 **NOTE:** This will very likely be split out into a separate repository at some point in the future.
 
 ## Windows on Windows


docs/administrator-docs/building.md
```diff
@@ -31,6 +31,9 @@ All package builds begin with the first two steps:
     `./build --list-platforms`
     `./build <platform> all`
 
+4. Run the executable:
+    `./Jellyfin.Server/bin/Release/netcoreapp<dotnetcore-version>/<platform>/jellyfin`
```
dkanada: I actually think the build scripts output to a specific directory, but I'm not sure which one. If this is the case we should update to the proper directory, otherwise this is fine. @joshuaboniface might know.

joshuaboniface:
Please note that those sets of trailing double spaces are for formatting and are required.

@dkanada is also correct; on completion of the scripts no binaries are present at that location - they get zipped up and then cleaned. The step should indeed reference where the resulring zip/tar archive is, then reference the (to be written) installation docs for the standalone binary packages.

DrPandemic:
Perfect, it seems that this PR don't align with what you are trying to do. I'll close it if it's the case.

In this case, maybe a page to use debug builds could be useful, what do you think? Also, normally, how do you run you dev builds?

DrPandemic:
Ok I just understood why I was confused about the packaging. The help from `./build` says that the output is going to be in `../jellyfin-build/<platform>` but it's actually in `../bin/<platform>`. So now this PR mentions where the final archive is going to be. I'll also do a quick PR on the build script to update the help message.

I think it could also be useful to add doc to know how to build and run debug builds. I would like writing it, but I'm not familiar enough with the dotnet core development process. Do you have pointers on how you are doing it?

anthonylavado:
AFAIK - the debut builds are useful if you are going to be debugging them with a code editor,  but there’s no other benefit to doing so. That’s the only difference between “debug” and “release” builds.

PR #76, opened 2019-04-28
Rework kodi addon instructions

mcarlton00:
Changing the instructions for installing the kodi addon to use the newly created repo

### DIFF
--- a/docs/clients.md
+++ b/docs/clients.md
@@ -132,12 +132,12 @@ The official Jellyfin Kodi plugin.
 1. Download the repository installer as a [zip file](https://repo.jellyfin.org/releases/client/kodi/repository.jellyfin.kodi.zip)
     * It will be saved as `repository.jellyfin.kodi.zip`
 2. Install the Jellyfin repository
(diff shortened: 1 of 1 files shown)


docs/clients.md
```diff
@@ -129,28 +129,33 @@ The official Jellyfin Kodi plugin.
 
 #### Installing
 
-1. Download the add-on as a [zip file](https://github.com/jellyfin/jellyfin-kodi/archive/master.zip)
-    * It will be saved as `jellyfin-kodi-master.zip`
-2. Install Jellyfin for Kodi
+1. Download the repository installer as a [zip file](https://repo.jellyfin.org/releases/client/kodi/repository.jellyfin.kodi.zip)
+    * It will be saved as `repository.jellyfin.kodi.zip`
+2. Install the Jellyfin repository
```
PrplHaz4: Might be good to specify that these steps are done from within Kodi (not Jellyfin)

mcarlton00: Made it a bit clearer to see that the installing is done on the kodi/client side, not the server side.

PR #81, opened 2019-05-31
Fix identation in block quotes in the Hardware Acceleration section

sbor23:


joshuaboniface:
The layout of this looks fine: 
Screenshot from 2019-05-31 09-25-27

ReadTheDocs has a very annoying and particular format for block quotes *within a list* - I've not seen the syntax you're using, but if it renders better in ReadTheDocs I'll accept it. Do you have a screenshot of this in action live on RTD?

sbor23:
Indeed it looks good now!
Had the patch lying around for a month or so and your recent changes made in https://github.com/jellyfin/jellyfin-docs/commit/5af7354c448f2a8cb7ddfae14a3129135c1233bd fix it already.

Sorry for bothering you :)

Also kinda weird github doesn't tell me right away that there would be merge conflicts when asked to create a PR...

joshuaboniface:
No problem, I was confused about that merge conflict too :) I was secretly hoping you found a new nicer format - I might test it out in some of my test docs and see!

PR #88, opened 2019-06-13
Layout changes and a small css tweak to hide the subnav

dkanada:


README.md
```diff
@@ -1,23 +1,28 @@
-# Jellyfin project documentation
+<h1 align="center">Jellyfin Documentation</h1>
```
joshuaboniface: I'm iffy about this - I feel like I tried doing this originally, and it was a failure on ReadTheDocs, but let's just try it.

PR #98, opened 2019-07-28
Modify issues docs to reference Fider

joshuaboniface:
Add links and description of using our Fider instance (https://features.jellyfin.org) to manage feature requests.


joshuaboniface:
I'd much prefer continuous if we can manage it, but I can tweak it if we can't do that.

EraYaN:
I'd hope that we can stop people for posting feature requests on the Github page and otherwise just close them, but the python script is easy enough to adapt I suppose.

PR #136, opened 2019-10-03
Add Windows installer docs and macOS app docs.

### LINKED ISSUE #85: Add clearer update instructions for Windows
Provide a set of instructions for Windows users on how to update their installations.

Referenced here:
https://www.reddit.com/r/jellyfin/comments/bysa53/jellyfin_1035_released/eqqz0hb/

EraYaN:
Closes #85

### DIFF
--- a/general/administration/installing.md
+++ b/general/administration/installing.md
@@ -63,14 +63,17 @@ An Unraid Docker template is available in the repository.
 
 A community project to deploy Jellyfin on Kubernetes-based platforms exists [at their repository](https://github.com/home-cluster/jellyfin-openshift). Any issues or feature requests related to deployment on Kubernetes-based platforms should be filed there.
 
-## Windows (x64)
+## Windows (x86/x64)
 
 Windows installers and builds in ZIP archive format are available [here](https://repo.jellyfin.org/releases/server/windows).
 
 [!WARNING]
-> If you installed a version prior to 10.4.0 using a powershell script, you will need to manually remove the service and uninstall the server. Also one might need to move the data files to the correct location, or point the installer at the old location.
+> If you installed a version prior to 10.4.0 using a powershell script, you will need to manually remove the service using the command `nssm remove Jellyfin` and uninstall the server by remove all the files manually. Also one might need to move the data files to the correct location, or point the installer at the old location.
 
-### Install using installer
+[!WARNING]
+> The 32-bit or x86 version is not recommended. `ffmpeg` and its video encoders generally perform better as a 64-bit executable due to the extra registers provided. This means that the 32-bit version of Jellyfin is deprecated.
+
+### Install using installer (x64)
 
 Only available for versions 10.4.0+.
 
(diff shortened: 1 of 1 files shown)


general/administration/installing.md
```diff
@@ -63,25 +63,38 @@ An Unraid Docker template is available in the repository.
 
 A community project to deploy Jellyfin on Kubernetes-based platforms exists [at their repository](https://github.com/home-cluster/jellyfin-openshift). Any issues or feature requests related to deployment on Kubernetes-based platforms should be filed there.
 
-## Portable Binaries
+## Windows (x64)
```
JustAMan: why only `x64`, did we drop 32-bit for Windows?

general/administration/installing.md
```diff
@@ -63,25 +63,38 @@ An Unraid Docker template is available in the repository.
...
+Windows installers and builds in ZIP archive format are available [here](https://repo.jellyfin.org/releases/server/windows).
 
-### Windows (x64/x86)
+[!WARNING]
+> If you installed a version prior to 10.4.0 using a powershell script, you will need to manually remove the service and uninstall the server. Also one might need to move the data files to the correct location, or point the installer at the old location.
 
-Windows builds in ZIP archive format are available [here](https://repo.jellyfin.org/releases/server/windows).
+### Install using installer
 
-### Install via script
```
JustAMan: I feel that this section should be retained for those who would want to work with pre-10.4.0 releases

general/administration/installing.md
EraYaN: There is no installer for x86. And I think we should drop it, nobody in their right mind should be using anything 32-bit, besides anything that is restricted to 32-bit will be to slow to run Jellyfin satisfactorily anyway.

general/administration/installing.md
```diff
@@ -116,15 +129,49 @@ Windows builds in ZIP archive format are available [here](https://repo.jellyfin.
...
+
+1. Download and extract the latest version
+1. Run `install.bat` or `install-jellyfin.ps1`
+1. (optional) Enter a custom install location (default is `%appdata%`)
+1. (optional) Enter a custom library/data location (default is `%localappdata%`) - Currently not working
+1. (optional) Select to install jellyfin as a service
+1. (optional) Import old library - Not compatible with newer Emby versions
+1. (optional) Create a shortcut on your desktop
+1. (optional) Run Jellyfin after installation
+1. Press `Install`
```
JustAMan: I feel that it's awkward when we say "when using installer please uninstall zip first" and at the same time we don't provide uninstall instructions for zip...

general/administration/installing.md
JustAMan: Can we note that 32-bit version is ZIP-only (and is deprecated but still available for now)?

general/administration/installing.md
EraYaN: Didn't I write that in the note?

JustAMan: I don't see any instructions on _how_ to uninstall stuff which was installed as zip

EraYaN:
I'll make a different PR to rewrite all those and fix all the other header formatting and the like.

PR #137, opened 2019-10-03
Rewrite download links and fix header levels

EraYaN:
Alright so the package repositories (ubuntu, debian, kodi)
Should probably go on a seperate page on the website.

### DIFF
--- a/general/administration/connectivity.md
+++ b/general/administration/connectivity.md
@@ -6,11 +6,11 @@ title: Connectivity
 
 # Connectivity
 
-### Connect
+## Connect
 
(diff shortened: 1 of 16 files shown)


joshuaboniface:
>Alright so the package repositories (ubuntu, debian, kodi) Should probably go on a seperate page on the website.

Do you mean in the docs, or on the Downloads page? For the latter I disagree fundamentally since these should be the only encouraged install methods for those platforms.

EraYaN:
Well I meant on the downloads page. And obviously it should be the default, but I just feel that the downloads page is anything but clear.

PR #142, opened 2019-10-08
Add information about Base URL

joshuaboniface:
Adds a section to the administrator docs regarding the Base URL setting, its function, and its current caveats for reference.

Related to jellyfin/jellyfin#1863


general/administration/connectivity.md
```diff
@@ -35,3 +35,19 @@ Since client auto-discover would break if this option were configurable, you can
...
+
+The entered value on the configuration page will be normalized to include a leading`/` if this is missing.
+
+This setting requires a server restart to change, in order to avoid invalidating existing paths until the administrator is ready to.
+
+There are three main caveats to this setting:
+
+1. When setting a new Base URL (i.e. from `/` to `/baseurl`) or changing a Base URL (i.e. from `/baseurl` to `/newbaseurl`), the Jellyfin web server will automatically handle redirects to avoid displaying users invalid pages. For instance, accessing a server with a Base URL of `/jellyfin` on the `/` path will automatically append the `/jellyfin` Base URL. However, entirely removing a Base URL (i.e. from `/baseurl` to `/`, an empty value in the configuration) will not - all URLs with the old Base URL path will become invalid and throw 404 errors. This should be kept in mind when removing an existing Base URL.
```
JustAMan: I don't remember the logic of "redirect on base url change", only when / is changed to something else... am I missing something?

general/administration/connectivity.md
```diff
@@ -35,3 +35,19 @@ Since client auto-discover would break if this option were configurable, you can
...
+
+2. Client applications generally, for now, do not handle the Base URL redirects explicitly. Therefore, for instance in the Android app, the `Host` setting *must* include the BaseURL as well (e.g. `http://myserver:8096/baseurl`), or the connection will fail.
```
JustAMan:
```suggestion
2. Client applications generally, for now, do not handle the Base URL redirects implicitly. Therefore, for instance in the Android app, the `Host` setting *must* include the BaseURL as well (e.g. `http://myserver:8096/baseurl`), or the connection will fail.
```
You had to be explicit by including base url, so I think this is the right word there

general/administration/connectivity.md
joshuaboniface: It will detect a change. So if you go from `/old` to `/new`, it will realize that `/old` no longer matches what it expects, and redirect back to the main page. It's only full removal that it can't redirect.

JustAMan: Whatever you say :)

PR #146, opened 2019-10-16
Add Security headers to Nginx remote proxy configuration.

iotku:
This commit adds the security headers for the example configuration of the nginx Reverse proxy for jellyfin-web. Related to https://github.com/jellyfin/jellyfin/issues/1885 https://github.com/jellyfin/jellyfin/issues/879

X-Frame-Options and X-XSS-Protection carry near-zero risk of breakage, however the Content Security Policy (CSP) has to be sufficiently relaxed as to not block functionality.

I would in particular prefer to remove 'unsafe-inline' for script-src and use sha hash for chromecast script, however inline scripts would trigger CSP blocking (seems to be mostly in the form of onclick() elements)

Additionally, by including a hash for the chromecast js 'unsafe-inline' is disabled so that is being left out of the CSP.

I'm not aware of any additional external js aside from the chromecast js, but by restricting script-src to 'self' external js needs to be explicitly included into the CSP.

### DIFF
--- a/general/administration/reverse-proxy.md
+++ b/general/administration/reverse-proxy.md
@@ -121,6 +121,14 @@ server {
 #    ssl_stapling on;
 #    ssl_stapling_verify on;
 #
+#    # Security / XSS Mitigation Headers
+#    add_header X-Frame-Options "SAMEORIGIN";
+#    add_header X-XSS-Protection "1; mode=block";
+#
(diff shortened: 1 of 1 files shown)


iotku:
Going to put this on the back-burner for a bit while I figure out why the CSP isn't cooperating with chrome.

PR #148, opened 2019-10-16
Add security headers to Nginx reverse proxy

iotku:
This commit adds the security headers for the example configuration of the nginx reverse proxy for jellyfin-web. Related to jellyfin/jellyfin#1885 jellyfin/jellyfin#879

X-Frame-Options, X-Content-Type-Options, and X-XSS-Protection carry near-zero risk of breakage, however the Content Security Policy (CSP) has to be sufficiently relaxed as to not block functionality.

I'm not aware of any additional external js aside from the chromecast js, but by restricting script-src to 'self' external js needs to be explicitly included into the CSP.

I would in prefer to remove 'unsafe-inline' for script-src and use a sha hash for chromecast script, however inline scripts would trigger CSP blocking (seems to be mostly in the form of onclick() elements)

Additionally, by including a hash for the chromecast js 'unsafe-inline' is disabled so that is being left out of the CSP.

### DIFF
--- a/general/administration/reverse-proxy.md
+++ b/general/administration/reverse-proxy.md
@@ -121,6 +121,15 @@ server {
 #    ssl_stapling on;
 #    ssl_stapling_verify on;
 #
+#    # Security / XSS Mitigation Headers
+#    add_header X-Frame-Options "SAMEORIGIN";
+#    add_header X-XSS-Protection "1; mode=block";
+#    add_header X-Content-Type-Options "nosniff";
+#
+#    # Content Security Policy
+#    # See: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
+#    add_header Content-Security-Policy "default-src https: data: blob:; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' https://www.gstatic.com/cv/js/sender/v1/cast_sender.js; worker-src 'self' blob:; connect-src 'self'; object-src 'none'";
+#
 #    location / {
 #        # Proxy main Jellyfin traffic
 #        proxy_pass http://SERVER_IP_ADDRESS:8096;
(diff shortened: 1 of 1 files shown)


iotku:
I'm still pretty 50/50 on the value of adding a CSP on the reverse proxy side of things especially if we need script-src 'unsafe-inline' still.

I don't think it hurts any, but a large amount of value in a CSP is added XSS protection 

https://csp.withgoogle.com/docs/why-csp.html

> Creating a policy which allows inline scripts (script-src 'unsafe-inline') or allows loading scripts from untrusted domains does not improve security because it doesn't protect the application from XSS. If an application cannot use a safe policy it will not benefit from CSP until all patterns incompatible with CSP are removed; as such, it is prudent to hold off on deploying a policy until the application can adopt strict CSP.

general/administration/reverse-proxy.md
```diff
@@ -121,6 +121,15 @@ server {
...
 #    ssl_stapling_verify on;
 #
+#    # Security / XSS Mitigation Headers
+#    add_header X-Frame-Options "SAMEORIGIN";
+#    add_header X-XSS-Protection "1; mode=block";
+#    add_header X-Content-Type-Options "nosniff";
+#
+#    # Content Security Policy
+#    # See: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
+#    add_header Content-Security-Policy "default-src https: data: blob:; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' https://www.gstatic.com/cv/js/sender/v1/cast_sender.js; worker-src 'self' blob:; connect-src 'self'; object-src 'none'";
```
JustAMan: can you clarify `cast_sender.js` part?

JustAMan: Thanks. Can we get a comment explaining why it's there added to the doc?

PR #149, opened 2019-10-17
xref:contrib-coding-style

stappersg:
Reporting a broken link (by breaking it further because my initial fix attempt failed)


dkanada:
Why not just remove the link entirely? Coding style will usually be dependent on the repository anyway, makes sense to just let each project handle it differently.

PR #153, opened 2019-10-30
added instructions for native plugin mode

brando56894:
Added from my how-to on [Reddit](https://www.reddit.com/r/jellyfin/comments/dolfvt/how_to_native_plugin_mode_for_kodi/) at the suggestion of mcarlton00


general/clients/clients.md
```diff
@@ -165,12 +165,24 @@ The official Jellyfin Kodi plugin.
...
 7. Once the full sync is done, you can browse and play your media through Kodi, and syncs will be done periodically in the background
 
+##### Native Mode Configuration
+
+The benefit of Native mode over Add-On mode is that nothing will be transcoded, currently Add-On mode doesn't support audio with more than 5.1 channels, which means no Dolby Atmos or DTS-X. Since no transcoding happens, files play instantly.
+
+1. You need to delete and re-add your libraries on your Jellyfin server in order to set the **Optional Network Path** to your share locations if you haven't already done so previously, SMB/Samba is preferred, and **must be in the Windows format** using backslashes, *even if both the client and server are using Unix-like OSes*, for example **\\\192.168.1.7\movies**, wait for the import to finish on the server.
+2. If using a Unix-like OS you must create a local user and Samba user named **guest** (with no password) on your server and allow guest only access to your shares, if using Windows this may already be done for you by default, the **guest** user is hardcoded into the plugin.
```
dkanada: Is this a Jellyfin limitation or an OS limitation? Linux has many different flavors so it might be specific to your distro if that's the case.

brando56894: You're welcome! Happy to help! I'm not sure what you're referring to limitation wise, but this should apply to all unix-like OSes (FreeBSD, OS X, Linux, etc...) where SMB/Samba is used. When I posted this on Reddit, someone said that the Jellyfin server will accept NFS paths as well, but I'm not sure if Kodi for Jellyfin supports NFS so I just stuck with SMB since that's what JFK mentions.

I couldn't test this out with Windows since I don't run anything with Windows in my apartment (Windows bad! hahaha)

PR #161, opened 2019-11-26
Updates around Kodi instructions

mcarlton00:
* Links directly to the repository download file, saving users a few clicks
* Updates kodi instructions to use the new default baseurl value to avoid confusion
* Changes the maintained status to "Active" (I think it's about time to update that)

### DIFF
--- a/general/clients/clients.md
+++ b/general/clients/clients.md
@@ -135,7 +135,7 @@ The official Jellyfin iOS app.
 
 The official Jellyfin Kodi plugin.
 
-**Status:** ⭐ Receiving patches
+**Status:** ⭐ Active
 
 **Links:**
 
@@ -143,7 +143,7 @@ The official Jellyfin Kodi plugin.
 
 #### Installing
 
-1. Download the repository installer [here](https://jellyfin.org/downloads/clients/#kodi-repository)
+1. Download the repository installer [here](https://repo.jellyfin.org/releases/client/kodi/repository.jellyfin.kodi.zip)
     * It will be saved as `repository.jellyfin.kodi.zip`
 2. Install the Jellyfin repository
     * Open Kodi and navigate to "Add-on Browser"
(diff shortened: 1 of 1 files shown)


general/clients/clients.md
```diff
@@ -135,15 +135,15 @@ The official Jellyfin iOS app.
...
+**Status:** ⭐ Active
 
 **Links:**
 
 * [GitHub](https://github.com/jellyfin/jellyfin-kodi)
 
 #### Installing
 
-1. Download the repository installer [here](https://jellyfin.org/downloads/clients/#kodi-repository)
+1. Download the repository installer [here](https://repo.jellyfin.org/releases/client/kodi/repository.jellyfin.kodi.zip)
```
dkanada: maybe we should just link to https://repo.jellyfin.org/releases/client/kodi

general/clients/clients.md
```diff
@@ -159,10 +159,11 @@ The official Jellyfin Kodi plugin.
 4. Within a few seconds you should be prompted for your server-details.
     * If a Jellyfin server is detected on your local network, it will displayed in the popup
     * If a Jellyfin server is not detected on your local network, select "Manually add server"
+        * Note that `/jellyfin` in these examples should match the `baseurl` field in the server (defaults to `/jellyfin`)
         * Enter the server name or IP address and the port number (default is 8096)
+            * Host: `192.168.1.10:8096/jellyfin`
         * If using SSL and a reverse proxy, enter the full URL scheme in the "Host" field
-            * Host: https://jellyfin.example.com
-            * Port: 443
+            * Host: `https://jellyfin.example.com/jellyfin`
```
dkanada: Since we have changed the default back to `/` I don't think we should leave `/jellyfin` in the documentation here.

general/clients/clients.md
mcarlton00: My thought was since we say "It'll download as ____", it's implying that is the actual download link.  Linking it to `/kodi` basically just adds another click for users.  Plus that download page is super ugly and unstyled after how nice the rest of the site looks

general/clients/clients.md
mcarlton00: Ah, didn't realized we'd changed the default back.  I'll change this part

PR #162, opened 2019-11-28
Update Traefik documentation

Artiume:
Updated Jellyfin docker information for traefik and jellyfin

### DIFF
--- a/general/administration/reverse-proxy.md
+++ b/general/administration/reverse-proxy.md
@@ -315,7 +315,7 @@ Add a job to cron so the certificate will be renewed automatically:
 
 ## Traefik (with docker-compose)
 
-Traefik is a modern HTTP reverse proxy and load balancer that makes deploying microservices easy. Traefik integrates with your existing infrastructure components (Docker, Swarm mode, Kubernetes, Marathon, Consul, Etcd, Rancher, Amazon ECS, ...) and configures itself automatically and dynamically. Pointing Traefik at your orchestrator should be the only configuration step you need. (https://traefik.io/)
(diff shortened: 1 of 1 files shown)


general/administration/reverse-proxy.md
```diff
@@ -327,31 +327,41 @@ docker-compose.yml :
...
 
 services:
-  proxy:
-    container_name: Traefik
-    image: traefik:alpine
-    command: --web --docker --docker.domain=${DOMAIN} --logLevel="INFO"
+  traefik:
+    container_name: traefik
+    hostname: traefik
+    image: traefik:traefik:v1.7.19
```
joshuaboniface: Should probably not specify a specific version in our docs, as this is likely to change frequently.

general/administration/reverse-proxy.md
```diff
@@ -327,31 +327,41 @@ docker-compose.yml :
...
       - 443:443
-    expose:
-      - 8080
+      - 8080:8080
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
       - ./traefik.toml:/traefik.toml
       - ./acme.json:/acme.json
-    restart: unless-stopped
+    restart: always
```
joshuaboniface: Unless-stopped would make more sense to me for an example, so that it could be easily shut down if required. My Docker-Compose knowledge is very limited though.

Artiume: You're right, Unless-stopped might be a better option for people. Unfortunately, due to traefik shifting v2 to their main branch, you have to specify which v1.7 version you want. Traefik v1 is deprecated at this point but I don't think a v2 guide would be appropriate for this.

joshuaboniface: Thanks for confirming the version part, I don't use Traefik myself so didn't realize they did that.

Artiume: fixed the restart policy, should be good to merge.

joshuaboniface:
Thanks!

Issue #124, opened 2019-09-24
Subfolders are being written to in docker (not really a bug) 


Uatschitchun:
This one is not really a bug, but a security problem!
As because of this:
https://github.com/docker/for-linux/issues/788

Adding a folder with submounts read only like:
`-v /media:/media:ro `
doesn't make mountpoints underneath /media, like /media/movies, read only, if movies is a mountpoint. 
This should be documented here:
https://jellyfin.readthedocs.io/en/latest/administrator-docs/installing/#docker-hub

I was wondering, as jellyfin wrote nfo files and posters to a bind mount that I thought is read only, although it isn't! 

I know of many people having their media added to jellyfin like this, who believe, their media is safe this way!

ThatNerdyPikachu:
@anthonylavado Does Jellyfin:
- ***ever*** write to the media files? (I assume not, just wondering :P)
- write to the folders containing the files on its own, without user configuration of a feature that does that?

Thanks!

anthonylavado:
@ThatNerdyPikachu 
1. No
2. No. As you said, you’d have to enable something, like NFO writing, or keeping images next to the media, etc.

JustAMan:
1. Not completely "No" :) there's that dready "Delete" button which an admin can push (and it _should_ not trigger automatically but bugs could happen as we all know)

dkanada:
Looks like a warning was added in the docker section so I'll close this.

PR #164, opened 2019-11-29
Update installing.md

Artiume:
Updated docker portions of documentation. Added a note so that an issue can be closed.

https://github.com/jellyfin/jellyfin-docs/issues/124

### DIFF
--- a/general/administration/installing.md
+++ b/general/administration/installing.md
@@ -40,14 +40,13 @@ version: "3"
 services:  
   jellyfin:  
     image: jellyfin/jellyfin
-    container_name: jellyfin
     network_mode: "host"  
     volumes:  
       - /path/to/config:/config  
       - /path/to/cache:/cache  
       - /path/to/media:/media  
-    restart: unless-stopped
 ```
+
 
 ### [Docker Hub](https://hub.docker.com/r/linuxserver/jellyfin) image maintained by LinuxServer.io
 
--- a/general/administration/installing.md
+++ b/general/administration/installing.md
@@ -53,35 +53,6 @@ services:
 
 <a href="https://hub.docker.com/r/linuxserver/jellyfin"><img alt="Docker Pull Count" src="https://img.shields.io/docker/pulls/linuxserver/jellyfin.svg"></a>
 
-Docker-compose example of the linuxserver.io image. 
-
-```
----
-version: "3"
-services:
-  jellyfin:
-    image: linuxserver/jellyfin
-    container_name: jellyfin
-    environment:
-      - PUID=1000
-      - PGID=1000
-      - TZ=Europe/London
-      - UMASK_SET=022 #optional
-    volumes:
-      - /path/to/library:/config
-      - /path/to/tvseries:/data/tvshows
-      - /path/to/movies:/data/movies
-      - /path for transcoding:/transcode #optional
-    ports:
-      - 8096:8096
-      - 8920:8920 #optional
-    devices:
-      - /dev/dri:/dev/dri #optional
-    restart: unless-stopped
-```
-
-Note: The linuxserver example does not call out host_mode for networking so a bridge is created. DLNA and HDHomeRun will not work in bridged mode. Add network_mode: host and remove the ports if host mode is wanted. If a reverse proxy is wanted in host_mode refer to the [reverse proxy](https://jellyfin.org/docs/general/administration/reverse-proxy.html) page.
-
 ### Unraid Docker
 
 An Unraid Docker template is available in the repository.
--- a/general/administration/installing.md
+++ b/general/administration/installing.md
@@ -10,11 +10,16 @@ The Jellyfin project and its contributors offer a number of pre-built binary pac
 
 ## Containers
 
+Note: There is currently an [issue](https://github.com/docker/for-linux/issues/788) for read-only mounts in Docker. If there are submounts within the main mount, the submounts are read-write capable. 
+
+Use host mode for networking in order to use DLNA or an HDHomeRun. 
+
 ### Official Docker Hub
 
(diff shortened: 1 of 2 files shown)


JustAMan:
cc @nvllsvm for review

general/administration/installing.md
```diff
@@ -27,25 +32,59 @@ The Jellyfin Docker image is available on [Docker Hub](https://hub.docker.com/r/
     `--volume /path/to/cache:/cache \`  
     `--volume /path/to/media:/media \`  
     `--net=host \`  
-    `jellyfin/jellyfin`  
+    `jellyfin/jellyfin:latest`
```
nvllsvm: This change is unnecessary; `latest` is always the default tag when no tag is specified.
Please remove.

general/administration/installing.md
```diff
@@ -27,25 +32,59 @@ The Jellyfin Docker image is available on [Docker Hub](https://hub.docker.com/r/
...
+    volumes:  
+      - /path/to/config:/config  
+      - /path/to/cache:/cache  
+      - /path/to/media:/media  
+    environment:
+      APP_UID: 1000
+      APP_UID: 1000
+      TZ: Europe/Paris
+      UMASK_SET: 022
+      GIDLIST=100 #A comma-separated list of additional GIDs to run emby as (default 2)#
```
nvllsvm: There is no such environment variable in the Jellyfin Docker image. If you need this capability, use Docker Compose spec v2.4 and the `group_add` feature. https://docs.docker.com/compose/compose-file/compose-file-v2/#group_add

Spec v3 removed it.

general/administration/installing.md
```diff
@@ -27,25 +32,59 @@ The Jellyfin Docker image is available on [Docker Hub](https://hub.docker.com/r/
...
+    restart: unless-stopped
 ```
 
 ### Docker Hub maintained by LinuxServer.io
 
 <a href="https://hub.docker.com/r/linuxserver/jellyfin"><img alt="Docker Pull Count" src="https://img.shields.io/docker/pulls/linuxserver/jellyfin.svg"></a>
 
+```
```
nvllsvm: There's no context to what this is. IMO - the LinuxServer.io section should be left as-is.

general/administration/installing.md
```diff
@@ -27,25 +32,59 @@ The Jellyfin Docker image is available on [Docker Hub](https://hub.docker.com/r/
...
 version: "3"  
 services:  
-    jellyfin:  
-      image: jellyfin/jellyfin  
-      network_mode: "host"  
-      volumes:  
-        - /path/to/config:/config  
-        - /path/to/cache:/cache  
-        - /path/to/media:/media  
+  jellyfin:
```
nvllsvm: None of the changes below are necessary or have no effect. The examples provided in the documentation should be as minimal as possible.

Please remove the below changes.

general/administration/installing.md
Artiume: I made the update as to be explicit to what version someone may be running rather than be implicit. Removing.

general/administration/installing.md
Artiume: what variables are supported? I tried running APP_UID since that's what the documentation had and it didn't work.

general/administration/installing.md
Artiume: It is the example provided by LinuxServer. I added it in there since it uses a bridge and I wanted to show an example. It is easier to use traefik using a bridge rather than host_mode because of this bug. 

https://github.com/containous/traefik/issues/5535

Let me know if you want it removed. Added more explanation instead.

Artiume:
updated changes after reviewing comments.

general/administration/installing.md
nvllsvm: There are no environment variables to modify the user and groups. The expectation is to use Docker-native features for that.

I'm not sure what env vars are supported, but I don't believe the UMASK_SET nor TZ are.

Artiume: Would you be opposed to PR's related to adding ENVs to the official dockerfile?

Artiume:
Overall, I'm trying to add more to the documentation due to issues such as this

https://github.com/jellyfin/jellyfin/issues/1322

I've posted a solution to the reverse proxy page and now trying to add to this page since a lot people are going to be using docker and don't fully understand docker all the time

anthonylavado:
@Artiume Minor conflict to resolve.
@nvllsvm Just looking for another review when you can 🙂

general/administration/installing.md
PrplHaz4: @Artiume the reason for using bridge mode is to make RP via Traefik easier - so I don't think it should be on this page.  This page should only address getting a fully working Jellyfin up and running as easily as possible.

Maybe it's fair to add that the linuxserver.io image does not support DLNA and HDHomeRun out of the box?

Artiume:
Removed lsio compose and moving the bridge warning to RP page.

Artiume:
This PR and the others created should be ready to merge. Request a review when feasible.

general/administration/installing.md
nvllsvm: Remove Remove the changes to this section.

Artiume: It fixed the indentation to be 2 spaces vice 4 spaces. I also added the restart policy since it was missing. Restarting docker would cause your container to not start due to the default restart policy being 'off'. 

If I'm missing something important, please give more detail to the issue at hand.

nvllsvm: This example should be as minimal as possible. Providing opinions like `name` and `restart` makes it more difficult to pick out the important parts.

Also - restarting is not always ideal ex. permission issue.

PR #166, opened 2019-11-30
Update to traefik

Artiume:
Due to the bug in traefik, a static route must be set in order to use network_mode: host in the yml. This should fix a lot of issues for people who want DLNA, HDHomeRun and FQDN together.

https://github.com/containous/traefik/issues/5559

### DIFF
--- a/general/administration/reverse-proxy.md
+++ b/general/administration/reverse-proxy.md
@@ -449,7 +449,9 @@ Congratulation, your RP is UP !
 
 For Jellyfin, just launch your Jellyfin server with this docker-compose `docker-compose up -d`.
 
-Note you must change the ${JELLYFIN_DOMAIN} for your domain, like jellyfin.mydomain.xyz for example. If using an HDHomeRun, use network_mode: host, remove the traefik network information for proper building of the yaml.
+Note: you must change the ${JELLYFIN_DOMAIN} for your domain, like jellyfin.mydomain.xyz for example. If using an HDHomeRun, use network_mode: host, remove the traefik network information for proper building of the yaml. 
+
+Note: Due to a [bug](https://github.com/containous/traefik/issues/5559) in traefik, you cannot dynamically route to containers, you must set a static route in your toml/file.
 
 ```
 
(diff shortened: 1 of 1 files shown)


anthonylavado:
@aritume Another conflict here now from the other PR

PrplHaz4:
Some feedback from me...I think a couple of these prevent it from running as-is.
- Static route is a bit confusing in this context - I think they should be called static rules, service or backend
- Traefik image should be `traefik:v1.7` (existing `traefik:traefik:v1.7.19` is invalid)
- network "traefik" is listed as external in both compose files, which means it needs to be preconfigured - it should probably be defined as a bridge network in traefik.yml and external in jellyfin.yml (or put them both in one compose file and define it there)
- indent on frontends.jellyfin is off
- remove basicAuth - it's commented, and looks like it needs some tweaking to actually be useful, and includes a cockpit config for some reason
- Note should read: "Note: Due to a bug in Traefik, you cannot dynamically route to containers running with host networking - you must set up static rules in your toml/file, and make sure that traefik can access the jellyfin ip."
- does a static rule that points the jellyfin backend to 172.18.0.1:8096 work?
- Maybe add a note at the beginning stating something like "DLNA and HDHomeRun LiveTV will not work with this configuration"

Artiume:
@PrplHaz4 
Fixed the issues. Yes, you can route it internally to the docker network but you have to make sure you set a static IP for the container.

PrplHaz4:
@Artiume I meant a static rule that points to the docker host, and jellyfin running with net=host.  A quick test now seems to work ok in that config, including DLNA - let me try to get a simple config together...

PrplHaz4:
@anthonylavado these changes look good to me - I'll try to submit a pr later that strips out a ton and hopefully gets some feedback (there's too many gotchas in here)

PrplHaz4:
@anthonylavado @Artiume I've created PR #168 to address the host networking piece and combine both services into a single yml, and simplify a bit (I hope).  DLNA and HDHR are both tested and working on my side.

Please review and provide feedback if you can.

Artiume:
Closing in favor of https://github.com/jellyfin/jellyfin-docs/pull/168

PR #168, opened 2019-12-05
Fix DLNA and HDHR for Traefik Docker compose

PrplHaz4:
Based on the changes proposed by @Artiume 

- Moved Traefik and Jellyfin to a single docker-compose.yml
- Removed some config items
- Added functional host networking so DLNA and LiveTV work
- Used static rules to point to docker host from Traefik network
- Changed docker-compose version to 3.5 (to support network name) - there are other workarounds but this is more copy/paste friendly

### DIFF
--- a/general/administration/reverse-proxy.md
+++ b/general/administration/reverse-proxy.md
@@ -15,7 +15,7 @@ Some popular options for reverse proxy systems are [Apache](https://httpd.apache
 When following this guide, be sure to replace the following variables with your information:
 
   * `DOMAIN_NAME` - Your public domain name to access Jellyfin on (e.g. jellyfin.example.com)
-  * `domain.tld` - The domain name Jellyfin services will run under (e.g. example.com)
+  * `example.com` - The domain name Jellyfin services will run under (e.g. example.com)
   * `SERVER_IP_ADDRESS` - The IP address of your Jellyfin server (if the reverse proxy is on the same server use 127.0.0.1)
 
 In addition, the examples are configured for use with LetsEncrypt certificates.  If you have a certificate from another source, change the ssl configuration from `/etc/letsencrypt/DOMAIN_NAME/` to the location of your certificate and key.
@@ -271,11 +271,11 @@ services:
       traefik.backend: traefik
       traefik.docker.network: traefik
       traefik.port: 8080
-      traefik.frontend.rule: Host:traefik.domain.tld,
+      traefik.frontend.rule: Host:traefik.example.com,
       traefik.frontend.entryPoints: https
       traefik.frontend.passHostHeader: "true"
       traefik.frontend.headers.SSLForceHost: "true"
-      traefik.frontend.headers.SSLHost: traefik.domain.tld
+      traefik.frontend.headers.SSLHost: traefik.example.com
       traefik.frontend.headers.SSLRedirect: "true"
       traefik.frontend.headers.browserXSSFilter: "true"
       traefik.frontend.headers.contentTypeNosniff: "true"
@@ -285,7 +285,7 @@ services:
       traefik.frontend.headers.STSPreload: "true"
       traefik.frontend.headers.customResponseHeaders: X-Robots-Tag:noindex,nofollow,nosnippet,noarchive,notranslate,noimageindex
       traefik.frontend.headers.frameDeny: "true"
-      traefik.frontend.headers.customFrameOptionsValue: 'allow-from https://domain.tld'
+      traefik.frontend.headers.customFrameOptionsValue: 'allow-from https://example.com'
     restart: unless-stopped
 
   jellyfin:
@@ -337,7 +337,7 @@ defaultEntryPoints = ["http", "https"]
 
 [acme]
 acmeLogging = true
-email = "user@domain.tld"
+email = "user@example.com"
 storage = "acme.json"
 entryPoint = "https"
   [acme.dnsChallenge]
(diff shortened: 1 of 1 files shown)


PrplHaz4:
> 
> 
> I think /acme.json is the right mount

I thought so too, but all the Traefik docs show it this way, and it seems to be working for me locally, so I'm going to leave it unless someone reports it being broken.

https://github.com/containous/traefik/blob/v1.7/docs/user-guide/docker-and-lets-encrypt.md

Artiume (approved):

JustAMan:
@PrplHaz4 please verify that your PR and https://github.com/jellyfin/jellyfin-docs/pull/170 don't clash over DLNA documentation :)

PrplHaz4:
@JustAMan no conflicts that I see - @Artiume and I talked about the changes prior to this PR...

Artiume (approved):

general/administration/reverse-proxy.md
```diff
@@ -8,13 +8,14 @@ title: Reverse Proxy
...
 
-Some popular options for reverse proxy systems are [Apache](https://httpd.apache.org/), [Haproxy](https://www.haproxy.com/), [Caddy](https://caddyserver.com/), and [Nginx](https://www.nginx.com/).
+Some popular options for reverse proxy systems are [Apache](https://httpd.apache.org/), [Haproxy](https://www.haproxy.com/), [Nginx](https://www.nginx.com/), [Caddy](https://caddyserver.com/) and [Traefik](https://traefik.io/).
 
 **Important:** In order for a reverse proxy to have the maximum benefit, you should have a publically routable IP address and a domain with DNS set up correctly.  These examples assume you want to run Jellyfin under a sub-domain (ie: jellyfin.example.com), but are easily adapted for the root domain if desired. Running Jellyfin in a subpath (example.com/jellyfin/) is supported by the Android and Web clients.
 
 When following this guide, be sure to replace the following variables with your information:
 
   * `DOMAIN_NAME` - Your public domain name to access Jellyfin on (e.g. jellyfin.example.com)
+  * `domain.tld` - The domain name Jellyfin services will run under (e.g. example.com)
```
oddstr13: Stick with `example.com/.org/.net` as per [RFC 2606](https://tools.ietf.org/html/rfc2606).

general/administration/reverse-proxy.md
```diff
@@ -239,117 +240,42 @@ DOMAIN_NAME {
 }
 ```
 
-## LetsEncrypt with Certbot
```
oddstr13: Is there any good reasons for changing the order of these sections?
As is, it makes the diff much harder to read.

oddstr13 (changes requested):

general/administration/reverse-proxy.md
Artiume: As the page currently is, I find it harder to read. The Let's Encypt section is compatible with several over RP's so it is sort of confusing to see Apache/Nginx configs again. having it at the bottom of the page helps isolate it. The entire RP page needs to be reworked to break the different sections into their own pages.

Artiume: As the page currently is, I find it harder to read. The Let's Encypt section is compatible with several over RP's so it is sort of confusing to see Apache/Nginx configs again. having it at the bottom of the page helps isolate it. The entire RP page needs to be reworked to break the different sections into their own pages.

Artiume: As the page currently is, I find it harder to read. The Let's Encypt section is compatible with several over RP's so it is sort of confusing to see Apache/Nginx configs again. having it at the bottom of the page helps isolate it. The entire RP page needs to be reworked to break the different sections into their own pages.

oddstr13: I agree that the different reverse proxy solutions need their own pages.
I think that should be done in a separate pr however, as the moving of the section makes reviewing this pr more difficult than it need be.

PrplHaz4:
Sorry @oddstr13 - my intention was to have atomic commits that could be reviewed independently (or thrown out) without making a bunch of PRs. The first commit is the real meat of this issue. 

I'll change to example.com and move reordering to another PR.

Agree they should all be on their own pages - but that will also require someone to dissect lets encrypt into nginx and apache pages (or DRY and put it on its own page).

PrplHaz4:
@oddstr13 I think I'm going back on undoing the ordering - the current order makes no sense and is confusing: 
- Apache
- HAProxy
- Nginx
- Caddy
- Lets Encrypt with Certbot
  - Apache
  - HAProxy
  - Nginx
- Traefik

My change does this (so at least the high level list of reverse proxies is consistent - certbot instructions for each proxy are moved to the bottom):
- Apache
- HAProxy
- Nginx
- Caddy
- Traefik
- Lets Encrypt with Certbot
  - Apache
  - HAProxy
  - Nginx

IMO, this is a reasonable stopgap solution until each RP is moved to its own page (and will hopefully help lots of the issues people are having with Traefik/--net=host.

general/administration/reverse-proxy.md
PrplHaz4: updated with commit 215dfe9

general/administration/reverse-proxy.md
PrplHaz4: Reviewing this commit on its own should help (the others are cosmetic): https://github.com/jellyfin/jellyfin-docs/pull/168/commits/a0a9e4887384ab3279fe9894bd8146a9a7b15b08

See my comments below about reordering - I really think it is better in this PR than the original.

oddstr13 (approved):
I don't know the specifics of docker-compose or Træfik, but looks good to me.

PR #186, opened 2020-01-05
Style improvements and client status updates

dkanada:


general/clients/installing-kodi.md
```diff
@@ -2,40 +2,41 @@
...
-        * If using SSL and a reverse proxy, enter the full URL scheme in the "Host" field
+        * If using SSL and a reverse proxy, enter the URL in the "Host" field
             * Host: `https://jellyfin.example.com`
     * Select user account and input password
-5. Once you're succesfully authenticated with the Jellyfin server, you'll be asked about your preferences for this device
+5. Once you're succesfully authenticated with the server, you'll be asked about your preferences for this device.
     * Choose your preferences for each options
       * Add-On mode requires no additional configuration
-      * To get Native mode to work, a few things must be done before the files are accessible, so skipping the initial import is necessary, skip the below steps and continue with the **Native Mode Configuration** steps
+      * To get Native mode to work skipping the initial import is necessary, so stop here and skip to **Native Mode Configuration**
```
EraYaN: That Native Mode Configuration should probably be an xref link.

general/server/users/adding-managing-users.md
```diff
@@ -1,9 +1,9 @@
 ---
 uid: server-users-adding-managing
```
EraYaN: Maybe change the uid to match the title and replace all xrefs to this uid too.

general/server/users/adding-managing-users.md
```diff
@@ -12,32 +12,36 @@ User management can be done under `Users` in the `Dashboard`. Here you can see y
...
 
 `Allow Live TV access` Unchecking this option will block the user's access to watch Live TV.
 
 `Allow Live TV recording management` Unchecking this option will block the user's access to set recording schedules.
 
 #### Media Playback
+
 `Allow media playback` Unchecking this option will block the user's access to media libraries, this does not include Live TV.
 
 > [!NOTE]
```
EraYaN: We should change all those `**Note:**` notes to this syntax as this renders like a nicely colored panel.

general/clients/installing-kodi.md
dkanada: This is a section in the same file though, so I don't think xref is needed here.

EraYaN: Fair enough I guess, although now they are close but they might be several page window sizes away at some point. The xref will point to the local anchor.

dkanada: I'll be doing more style work in future PRs, we can work on proper links then.

PR #199, opened 2020-01-14
live tv url bad formatting

Artiume:


JustAMan:
What does this change?.. why an extra line is now suddenly so important?

Artiume:
Opening it in Chrome/chromium doesnt break the url onto a new line while Firefox respects the newline.

PR #217, opened 2020-01-31
Remove more blank index pages

dkanada:


EraYaN:
This does make the name in the header "Quick Start" from "General"

dkanada:
Hmmm that's not ideal. Maybe I should just expand on the general page or something.

PR #222, opened 2020-02-04
dlna base url note

Artiume:


general/administration/connectivity.md
```diff
@@ -33,6 +33,8 @@ If DLNA fails to bind properly, `[ERR] Failed to bind to port 1900: "Address alr
 
 Setting `Alive message interval (seconds)` to 30 seconds also appears to help discovery for some clients.
 
+If a Base URL is set, try removing the base URL.
```
dkanada:
```suggestion
If a base URL is set, try removing the base URL.
```

PrplHaz4:
@artiume I think this should mention restarting the server after changing the base URL:

`If a base URL is set, try removing the base URL and restarting the Jellyfin service.`

Artiume:
Done

PR #223, opened 2020-02-06
Update CSS page

Artiume:


general/clients/skins.md
```diff
@@ -5,7 +5,7 @@ title: CSS Customization
 
 # CSS Customization
 
-[Custom CSS](https://www.w3schools.com/cssref/css3_browsersupport.asp) provides interface customization such as changing colors or modifying behavior. Here are a collection of community scripts available. Currently, the CSS modifications only work on the web client. To implement these changes, go to Dashboard > General > Custom CSS.
+[Custom CSS](https://www.w3schools.com/cssref/css3_browsersupport.asp) provides interface customization such as changing colors or modifying behavior. Here are a collection of community scripts available. Currently, the CSS modifications only work on the web client. The code will apply the code in order that it is written so code can override previously stated custom CSS. To learn more see [CSS Specificity](https://www.w3schools.com/css/css_specificity.asp). To implement these changes, go to Dashboard > General > Custom CSS.
```
thornbill: We should link to MDN instead of w3schools. https://developer.mozilla.org/en-US/docs/Web/CSS

Artiume: Done, changed the main link. Left the Specificity link as is since it's a good read.

thornbill:
W3schools had a pretty bad reputation and that’s stuck with me over the years. (See https://meta.stackoverflow.com/questions/280478/why-not-w3schools-com for more discussion around this.) It sounds like they may have improved but I still stick with MDN. 😄

Here is an article on MDN about specificity. https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

Feel free to use whichever you think explains it best though!

Artiume:
@thornbill thanks for the stack link, it makes sense why i had instantly found their links first. I've updated the page.

general/clients/skins.md
```diff
@@ -33,32 +33,54 @@ Green Mod:
...
+`.headerRight { color: yellow; }`
 
-![image](https://user-images.githubusercontent.com/20715731/73949397-5f6f2500-48c8-11ea-9eca-bc1eb61f1281.png)
+![image](https://user-images.githubusercontent.com/20715731/73962770-0d84ca00-48dd-11ea-9b50-563f8b4aa33b.png)
+
+#### Console Panel
+
+`.mainDrawer-scrollContainer { color: yellow; }`
+
+![image](https://user-images.githubusercontent.com/20715731/73963663-c13a8980-48de-11ea-9342-d1e89690e7b1.png)
```
dkanada: We should probably store these locally rather than on GitHub.

PR #224, opened 2020-02-07
add [file] directive to traefik config

PrplHaz4:
won't work without it

### DIFF
--- a/general/administration/reverse-proxy.md
+++ b/general/administration/reverse-proxy.md
@@ -362,6 +362,7 @@ domain = "example.com"
 network = "traefik"
 exposedbydefault = false
 
+[file]
 [backends]
   [backends.backend-jellyfin]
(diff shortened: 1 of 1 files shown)


Artiume:
Did you test it? I have an external file so I never thought about the fact that might be needed

PrplHaz4:
> 
> 
> Did you test it? I have an external file so I never thought about the fact that might be needed

yup, that's what's running on my test server - the `[file]` directive is there - probably got clipped when I tried to trim things moving it into the docs...

Official Traefik doc on using `[file]`: https://docs.traefik.io/v1.7/configuration/backends/file/

huuub:
Could anybody describe what is the purpose and how do I translate it to traefik2?

Artiume:
This is all for v1.7, there are currently no docs for v2 and they should be written from scratch

PR #225, opened 2020-02-07
Update Traefik's Dashboard Security

Artiume:


PrplHaz4:
I actually took some of this out when we re-did it.

I don't think this belongs in the Jellyfin docs - there's tons of documentation about securing Traefik, and we shouldn't be attempting to provide hardening for a publicly accessible service (and any number of config tweaks).

Artiume:
Either we remove the dashboard or we harden. I don't want to be in-between.

PrplHaz4:
> 
> 
> Either we remove the dashboard or we harden. I don't want to be in-between.

By default, Jellyfin passwords are sent in the clear, so I don't think that hard line is followed anywhere else...I don't think this is in between - it is not secure - same as the default install of Jellyfin.  I'm ok with removing it too, but likely the first step for troubleshooting will be enabling it.

Artiume:
Jellyfin's open passwords are typically on the same network since dlna can't accept session cookies, it has to do plain text. If it's on the internet, it should be protected by the TLS tunnel. Having the traefik dashboard publicly exposed imo is a hundred times worse than anything jellyfin is doing. Also, it's your reverse proxy, having potential multiple services that can be exposed, not just the single jellyfin service which shouldn't have any elevated privileges.

PrplHaz4:
> 
> 
> Jellyfin's open passwords are typically on the same network since dlna can't accept session cookies, it has to do plain text. If it's on the internet, it should be protected by the TLS tunnel. Having the traefik dashboard publicly exposed imo is a hundred times worse than anything jellyfin is doing. Also, it's your reverse proxy, having potential multiple services that can be exposed, not just the single jellyfin service which shouldn't have any elevated privileges.

I completely agree - the only problem I have is the assumption that it will be internet facing (my RP is not) - if so, I say we remove the dashboard access altogether to keep attack surface lower.

Artiume:
So here's the issue. If we disable the dashboard, I'm ripping out the labels as well. There's zero need for them. I think it's just easier to merge these few lines of code instead of making the traefik yml a shadow of its former self.

Artiume:
Hey @PrplHaz4 so I really want to keep the security feature, here's what I got. 

> I don't think this belongs in the Jellyfin docs - there's tons of documentation about securing Traefik, and we shouldn't be attempting to provide hardening for a publicly accessible service (and any number of config tweaks).

Thinking about it, I disagree with this. So 2FA auth for Portainer is a paid feature, I was upset about this because why should security be behind a paid wall. I find any form of security should be inherently free and provided. With regards to Jellyfin, I feel that anything we provide, we should absolutely ensure the security of our users. 

Also, for config tweaks, we're supposed to be the experts. All day long we have people coming in because they tweaked their system one way or another, and I've been slowly adding those tweaking nuisances to the docs. I'm going to be adding rclone/mergerfs configs to the docs (which I want to rename to wiki haha) because I'm tired of answering the same question weekly :P

What I would like to go ahead and do is make a few big changes. I want to split Networking out of Administration and do a proper split for the Reverse Proxy page for each associated proxy so it's not overwhelmingly big. We can pull DLNA, subnetting of Docker and the other various things into the networking page because it's currently split across several areas. Thoughts?

PrplHaz4:
If you're comfortable that this config is secure enough to expose to the internet I'm fine with it, but I know I'm not qualified to make that judgement and am always hesitant about telling people to expose something when it's likely they may not know what good looks like. 

It would be great to have them on individual pages and definitely help with documenting more detailed configs.

Artiume:
BasicAuth isn't perfect, but it's secure enough for home use. 

I created this to start the migration. https://github.com/jellyfin/jellyfin-docs/pull/226

PrplHaz4:
I've got no problems with basic auth. The SSL and Jellyfin configs now need to be confirmed too though since we have to assume it's being exposed.

Artiume:
It's as best as we can get it for v1.7, I hardened the SSL when I first came around.

https://github.com/jellyfin/jellyfin-docs/commit/5669a008370235b709488d822e71431a42183012#diff-05285db094d737aefb25629df81783ef

PR #230, opened 2020-02-09
Update nginx subpath with baseurl instructions

mcarlton00:
In the event a user wants to use a subpath, we should recommend using the baseurl so the application is aware that it's running at `/jellyfin` instead of the root

### DIFF
--- a/general/administration/reverse-proxy.md
+++ b/general/administration/reverse-proxy.md
@@ -157,7 +157,7 @@ server {
 #    }
 #    location /socket {
 #        # Proxy Jellyfin Websockets traffic
-#        proxy_pass http://SERVER_IP_ADDRESS:8096/;
+#        proxy_pass http://SERVER_IP_ADDRESS:8096/socket;
 #        proxy_http_version 1.1;
 #        proxy_set_header Upgrade $http_upgrade;
 #        proxy_set_header Connection "upgrade";
--- a/general/administration/reverse-proxy.md
+++ b/general/administration/reverse-proxy.md
@@ -173,8 +173,9 @@ server {
 
 ## Nginx with subpath
 
-When connecting to server from a client application, enter ``http(s)://DOMAIN_NAME/jellyfin`` in the address field, and **clear the port field**.
-Not all clients may handle this properly, but this is currently working for the web and Android clients.
+When connecting to server from a client application, enter ``http(s)://DOMAIN_NAME/jellyfin`` in the address field.
+
+Set the base URL field in the Jellyfin server.  This can be done by navigating to the Admin Dashboard -> Networking -> Base URL in the Jellyfin Web UI.  Fill in this box with `/jellyfin` and click Save.  The server will need to be restarted before this change takes effect.
 
 ```
 # Jellyfin hosted on http(s)://DOMAIN_NAME/jellyfin
@@ -197,13 +198,8 @@ server {
 
     # Jellyfin
     location /jellyfin {
-        return 302 $scheme://$host/jellyfin/;
-    }
-
-    location /jellyfin/ {
         # Proxy main Jellyfin traffic
-        # The / at the end is significant.
-        proxy_pass http://SERVER_IP_ADDRESS:8096/;
+        proxy_pass http://SERVER_IP_ADDRESS:8096;
 
         proxy_pass_request_headers on;
 
--- a/general/administration/reverse-proxy.md
+++ b/general/administration/reverse-proxy.md
@@ -197,7 +197,7 @@ server {
     }
 
     # Jellyfin
-    location /jellyfin {
+    location /jellyfin/ {
         # Proxy main Jellyfin traffic
         proxy_pass http://SERVER_IP_ADDRESS:8096/jellyfin/;
 
--- a/general/administration/reverse-proxy.md
+++ b/general/administration/reverse-proxy.md
@@ -197,7 +197,7 @@ server {
     }
 
     # Jellyfin
-    location /jellyfin/ {
+    location /jellyfin {
         return 302 $scheme://$host/jellyfin/;
     }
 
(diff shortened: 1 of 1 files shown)


general/administration/reverse-proxy.md
```diff
@@ -197,13 +198,8 @@ server {
...
     # Jellyfin
     location /jellyfin {
-        return 302 $scheme://$host/jellyfin/;
-    }
-
-    location /jellyfin/ {
         # Proxy main Jellyfin traffic
-        # The / at the end is significant.
-        proxy_pass http://SERVER_IP_ADDRESS:8096/;
+        proxy_pass http://SERVER_IP_ADDRESS:8096;
```
Artiume: Do we want to remove the trailing slash?

https://github.com/GrrrDog/weird_proxies/wiki/nginx#proxy_pass

mcarlton00: Not if we're using the base url.  A trailing slash more or less strips part of the passed URL string out when passing it to the backend.  Can't check right now, but if I remember right then it strips out the location block it's in.  So
```
location /jellyfin {
    proxy_pass http://$IP:8096/
```
would only pass `http://$IP:8096/web/index.html` back.

If we have a baseurl of `/jellyfin` set, then we want to make sure that the full `http://domain.com/jellyfin/whatever` gets passed back

Artiume: So if the internal service as a Path, I know you can handle it with traefik. I imagine nginx also does too (since nginx is ya kno, better lol).

Does it fail to parse correctly when jellyfin doesn't have a base url set? Or is the lack of closing slash only needed when a base url is set?

mcarlton00: What's in the repo now should be pretty close to working for no base url but running JF in a sub-path (I admittedly didn't check), but it presents other problems.  For example, Jellyfin then isn't aware that it's running behind a sub-path so things like autodiscovery won't work.  Jellyfin will announce "hey I'm at http://192.168.1.10" instead of `http://192.168.1.10/jellyfin`.  If we're going to provide documentation to use a sub-path, we should probably recommend the base url approach to have a more complete solution.

Artiume: What I mean is with the server at http://192.168.1.10/jellyfin can you assign the proxy pass to appropriately pass the entire line include the Path variable. I've had to do this with pihole to normalize the FQDN url

I think this is sort of what I mean? 
https://serverfault.com/questions/527907/redirect-a-subpath-to-a-external-host-with-nginx

mcarlton00:
We could if you want, but the real reason it was there before is because the previous config straight up wouldn't work without it.  Now it works both with and without

general/administration/reverse-proxy.md
```diff
@@ -197,13 +198,8 @@ server {
...
     # Jellyfin
     location /jellyfin {
-        return 302 $scheme://$host/jellyfin/;
-    }
-
-    location /jellyfin/ {
         # Proxy main Jellyfin traffic
-        # The / at the end is significant.
-        proxy_pass http://SERVER_IP_ADDRESS:8096/;
+        proxy_pass http://SERVER_IP_ADDRESS:8096/jellyfin/;
```
JustAMan: This doesn't feel correct. It can have unintended side effects like passing to a JF server a url like `DOMAIN/jellyfin2/foo` -> `INTERNAL/jellyfin/2/foo`

general/administration/reverse-proxy.md
```diff
@@ -157,7 +157,7 @@ server {
 #    }
 #    location /socket {
 #        # Proxy Jellyfin Websockets traffic
-#        proxy_pass http://SERVER_IP_ADDRESS:8096;
+#        proxy_pass http://SERVER_IP_ADDRESS:8096/;
```
JustAMan: this feels wrong, it would strip `/socket` from the passed URI before proxying it to JF

general/administration/reverse-proxy.md
```diff
@@ -144,7 +144,7 @@ server {
 #
 #    location / {
 #        # Proxy main Jellyfin traffic
-#        proxy_pass http://SERVER_IP_ADDRESS:8096;
+#        proxy_pass http://SERVER_IP_ADDRESS:8096/;
```
JustAMan: this feels useless a bit, as it basically makes `nginx` strip leading slash and then append one in the beginning... the only effect it has is normalizing the URI before passing further, don't think we want it here

Artiume:
So I pushed to have the slashes  put in to minimize xss injection

https://www.acunetix.com/blog/articles/a-fresh-look-on-reverse-proxy-related-attacks/

See the nginx proxy pass section. 

Mcarlton00 verified operations, would you mind testing out the config as well before we go changing it?

JustAMan:
Okay, I can live with slashes being put. But they should be put correctly to not break when you don't expect it :)

Artiume:
I don't know nginx so I won't be able to verify functionality.

Artiume:
I would like to push it as is so i can push the network restructure. if there ends up being any more issues, someone will come point them out to us lol

general/administration/reverse-proxy.md
JustAMan:
```suggestion
#        proxy_pass http://SERVER_IP_ADDRESS:8096/socket;
```

general/administration/reverse-proxy.md
```diff
@@ -196,14 +197,9 @@ server {
     }
 
     # Jellyfin
-    location /jellyfin {
-        return 302 $scheme://$host/jellyfin/;
-    }
```
JustAMan: don't remove this pls

general/administration/reverse-proxy.md
```diff
@@ -197,8 +197,16 @@ server {
     }
 
     # Jellyfin
+    location /jellyfin/ {
```
JustAMan:
```suggestion
    location /jellyfin {
```
this is important, these two locations are different!

Artiume: fixed

PR #242, opened 2020-02-20
Add matrix android channel to "Getting Help"

nielsvanvelzen:
Also updated the dev-client as it still used an old name


dkanada:
@nielsvanvelzen we'll have to wait until the CI gets fixed, not sure what's wrong.

PR #245, opened 2020-02-20
kill jellyfin with pidof inside docker

PrplHaz4:
(obviously) the PID of jellyfin can differ - this will make it more likely to work....


PrplHaz4:
> 
> 
> Did you test that? I'm not completely sure the binary is called `jellyfin` :)

I was surprised too but it seems to work for me - I think this was changed [here ](https://github.com/jellyfin/jellyfin/commit/f1167ace7e510c459341ab60337edc5385bcdc92#diff-3254677a7917c6c01f55212f86c57fbf) but since the nightly docker build was failing for so long never saw it.
```
ENTRYPOINT ["./jellyfin/jellyfin", \
    "--datadir", "/config", \
    "--cachedir", "/cache", \
    "--ffmpeg", "/usr/local/bin/ffmpeg"]
```

PR #247, opened 2020-02-20
Force generated newline on command line examples

wwwutz:
Add two trailing spaces behind command line examples to force
generating a newline a.k.a `<BR>`. This prevents joining the
commands in one line and confuse the casual cut&paste user.

### DIFF
--- a/general/contributing/development.md
+++ b/general/contributing/development.md
@@ -34,7 +34,7 @@ The first step is to set up a copy of the Git repository of the project you want
 1. On GitHub, "Fork" the Jellyfin repository you wish to contribute to, to your own user account using the "Fork" button in the relevant repository.
 
 1. Clone your fork to your local machine and enter the directory:  
-    `git clone git@github.com:yourusername/projectname.git`
+    `git clone git@github.com:yourusername/projectname.git`  
     `cd projectname/`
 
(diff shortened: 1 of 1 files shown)


wwwutz:
here's an example: Source `.md` with no trailing spaces:

screenshot-#247-source-no-linebreak

results in no line breaking as seen here:

screenshot-#247-no-linebreak

Whereas Source `.md` with treailing spaces:

screenshot-#247-source-with-linebreak-by-trailing-spaces

results in broken lines:

screenshot-#247-with-linebreak-by-trailing-spaces


That's what this PR is about 8-)

dkanada:
I would much rather use explicit break tags, but since the rest of the docs use spaces this should be good for now. This is one of my least favorite markdown rules, whitespace is so annoying to use.

wwwutz:
Just decide how you'd prefer it. I'd volunteer to fix it everywhere.

dkanada:
I would say that in most cases several lines of code should be in a code block instead. If several individual inline statements are required then a break tag is the way to go.

dkanada:
That way we can update the editor config to disallow trailing whitespace.

wwwutz:
Ok, I'll hunt down trailing double spaces and replace them by `<br/>` and add the tag, where it's needed.

wwwutz:
obsolteted by #248

PR #248, opened 2020-02-21
Fix line breaks and remove trailing spaces

wwwutz:
- replace double space line breaks with `<br/>` tag
- remove trailing spaces
- replace double-backticks when no single backtick is quoted
- convert ^M endings to LF

PR #250, opened 2020-02-21
Refactor some files

dkanada:


wwwutz:
I introduced some bad removal of linebreaks in #248. I'll fix them when #250 is merged.

PR #252, opened 2020-02-23
Add consistent code quoting

wwwutz:
- Move shell commands in a triple-backtick section.
- fix autonumbering

### DIFF
--- a/general/administration/building.md
+++ b/general/administration/building.md
@@ -12,29 +12,37 @@ Jellyfin supports several methods of building for different platforms and instru
 All package builds begin with these two steps:
 
 1. Clone the repository:
-    `git clone https://github.com/jellyfin/jellyfin.git`<br/>
-    `cd jellyfin`
+    ```sh
+    git clone https://github.com/jellyfin/jellyfin.git
+    cd jellyfin
+    ```
 
 2. Initialize the submodules:
-    `git submodule update --init`
+    ```sh
+    git submodule update --init
+    ```
 
 ## Docker
 
 3. Build the Docker image:
-    `docker build -t $USERNAME/jellyfin .`
+    ```sh
+    docker build -t $USERNAME/jellyfin .
+    ```
 
 4. Run the container:
-    `docker run -d -p 8096:8096 $USERNAME/jellyfin`
+    ```sh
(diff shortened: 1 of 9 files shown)


wwwutz:
Before:
fix-code-before

After:
fix-code-after

general/administration/building.md
```diff
@@ -12,29 +12,37 @@ Jellyfin supports several methods of building for different platforms and instru
 All package builds begin with these two steps:
 
 1. Clone the repository:
-    `git clone https://github.com/jellyfin/jellyfin.git`<br/>
-    `cd jellyfin`
+    ```sh
+    git clone https://github.com/jellyfin/jellyfin.git
+    cd jellyfin
+    ```
```
dkanada: What will the indentation look like in DocFX? Will code blocks even respect indentation?

wwwutz: yes, they will. I fed i to `docfx` through a docker image and it did what I expected.

wwwutz: docfxindent

dkanada: Since we're making changes to these code blocks, can we add newlines before and after as well?

wwwutz:
Yes we can add more newlines, as long as the code quote is indented:

No empty newline behind number:

num-no-newline

With empty newlines behind number:

num-with-newline png

BUT we have to have an empty newline  between numbered paragraphs, otherwise this will happen:

num-no-newline-at-all

This is why I added them in `general/clients/installing-mopidy.md`. Was quite a struggle to find out why code quoting didn't work there.

dkanada:
Alright cool, let's add newlines when possible then.

wwwutz:
anything I should change/edit/fix ?

Artiume:
Looks good, thanks for all the effort. Not for this PR, but what are your thoughts of making the guide using pipenv instead of pip, that way mopidy doesn't have to monkey with the system pip? I haven't used mopidy so I don't know if it'd have any effect on it.

wwwutz:
> your thoughts of making the guide using pipenv instead of pip,

none. not even one neuron gets triggered.

> I haven't used mopidy ...

me neither. I might take a look at it, but I doubt  I'll dig into that deeper.

Artiume:
Haha, not a fan?

PR #256, opened 2020-03-01
Update Kodi native path mapping to include user/pass

netpok:
Adds information how to pass a username/password combination for native playback.

Related: jellyfin/jellyfin-kodi#182


general/clients/installing-kodi.md
```diff
@@ -43,7 +43,7 @@ title: Installing Kodi
...
 
 1. You need to delete and re-add your libraries on your Jellyfin server in order to set the **Optional Network Path** to your share locations if you haven't already done so previously, SMB/Samba is preferred, and **must be in the Windows format** using backslashes, *even if both the client and server are using Unix-like OSes*, for example **\\\192.168.1.7\movies**, wait for the import to finish on the server.
-2. If using a Unix-like OS you must create a local user and Samba user named **guest** (with no password) on your server and allow guest only access to your shares, if using Windows this may already be done for you by default, the **guest** user is hardcoded into the plugin.
+2. If using a Unix-like OS you must create a local user and Samba user named **guest** (with no password) on your server and allow guest only access to your shares, if using Windows this may already be done for you by default. Alternatively an user password combination can be provided in the the **\\\\username:password@host\share** format.
```
JustAMan:
```suggestion
2. If using a Unix-like OS you must create a local user and Samba user named **guest** (with no password) on your server and allow guest only access to your shares, if using Windows this may already be done for you by default. Alternatively, a user-password combination can be provided in the the **\\\\username:password@host\share** format.
```

netpok: Thanks for the suggestion, English is not my native language so I appreciate all help.

PR #260, opened 2020-03-04
Rework Kodi page

Artiume:


general/toc.yml
```diff
@@ -2,7 +2,7 @@
   items:
   - uid: clients-css-customization
   - uid: clients-codec-support
-  - uid: clients-installing-kodi
+  - uid: clients-kodi
   - uid: clients-installing-mopidy
```
dkanada: You might want to rename this one as well in the future.

Artiume: Yeah, definitely. I could also add a page for the mpv shim as well.

PR #262, opened 2020-03-05
Small code quote fix

wwwutz:
- Move shell commands in a triple-backtick section.


mcarlton00:
Thanks for the fix.

Also, forgot I don't have merge rights over here.  Whoops

PR #269, opened 2020-03-10
add basic auth to traefik

Artiume:
http://www.duckdns.org/

https://github.com/linuxserver/docker-duckdns/


Artiume:
I'm adding it because of this

https://www.reddit.com/r/jellyfin/comments/fgdj6e/server_is_remotely_accessible_through_the_web_but/fk40i05

No matter what we do, people will expect to be capable of streaming remotely. The best we can do is educate them enough to protect themselves. Duckdns is a great way to have a single domain FQDN that is Foss

dkanada:
DuckDNS should be removed here as well I guess but the rest looks fine.

dkanada:
Now the commit history is too long 😅 have to squash this one and the other one first since I can't do it on GitHub anymore.

Artiume:
It's fixed. I dropped the compose to v2 since that's necessary for group_add (Hwa). I need to go through and add it to the other composes

PrplHaz4:
> 
> 
> It's fixed. I dropped the compose to v2 since that's necessary for group_add (Hwa). I need to go through and add it to the other composes

have you tested that it doesn't break networking?  i thought there was a problem where compose wouldn't use the pre-defined network (which is why it needed v 3.5 and the named network in docker-compose.yml) instead it creates a new bridge network (called traefik_traefik) - IIRC it was a while ago...

general/networking/traefik.md
```diff
@@ -57,6 +66,7 @@ services:
       traefik.frontend.headers.customResponseHeaders: X-Robots-Tag:noindex,nofollow,nosnippet,noarchive,notranslate,noimageindex
       traefik.frontend.headers.frameDeny: "true"
       traefik.frontend.headers.customFrameOptionsValue: 'allow-from https://example.com'
+#      traefik.frontend.auth.basic.users: xxx:xxx
```
JustAMan, line 69: Note that this would break some clients that doesn't know how to deal with `http basic auth` like AndroidTV. Should probably be mentioned somewhere.

Artiume: Not in this instance. This is basic auth for traefik's dashboard, not jellyfin. I haven't added auth rules to jellyfin yet specifically for that reason.

PR #270, opened 2020-03-11
Update the nginx config

aidanharris:
* Use a variable to refer to the backend server
* Add a location block for /web so /web/#!/ works instead of needing to
  go to /web/index.html/#!/. Without this the following error is
displayed in the browser:

> Value cannot be null. (Parameter 'path')


general/networking/nginx.md
```diff
@@ -107,7 +133,7 @@ server {
         # The / at the end is significant.
         # https://www.acunetix.com/blog/articles/a-fresh-look-on-reverse-proxy-related-attacks/
 
-        proxy_pass http://SERVER_IP_ADDRESS:8096/jellyfin/;
+        proxy_pass http://$jellyfin:8096/jellyfin/;
```
Artiume: @dkanada thoughts? we've avoided using environmentals for aliases. I was for it because I'm accustomed to them due to docker, but non-docker and windows people won't necessarily know what they mean.

dkanada: I have no opinion besides picking a standard, since we currently use a wide range of methods for situations like this.

general/networking/nginx.md
```diff
@@ -10,65 +10,85 @@ title: Nginx Reverse Proxy
...
+        proxy_set_header X-Forwarded-Host $http_host;
+
+        # Disable buffering when the nginx proxy gets very resource heavy upon streaming
+        proxy_buffering off;
+    }
+
+    # location block for /web - This is purely for aesthetics so /web/#!/ works instead of having to go to /web/index.html/#!/
+    location ~ ^/web/$ {
+        # Proxy main Jellyfin traffic
+        proxy_pass http://$jellyfin:8096/web/index.html;
```
Artiume:
```suggestion
        proxy_pass http://$jellyfin:8096/web/index.html/;
```

Can you test this fix? On a lot of these, we should be adding terminating slashes.

general/networking/nginx.md
```diff
@@ -10,65 +10,85 @@ title: Nginx Reverse Proxy
...
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_set_header X-Forwarded-Protocol $scheme;
+        proxy_set_header X-Forwarded-Host $http_host;
+    }
+
+    location /socket {
+        # Proxy Jellyfin Websockets traffic
+        proxy_pass http://$jellyfin:8096/socket;
```
Artiume: would a slash work here as well?

Artiume:
@mcarlton00 would you like to give this a once over?

aidanharris:
There's a few more changes I need to make so don't merge this yet. I think adding the trailing slashes may have broken something, and we also need a `listen [::]:80` and `listen [::]:443` otherwise IPV6 doesn't work.

aidanharris:
@mcarlton00 This should be ready for a review now. I haven't tested the "Nginx with Subpath" because I use Jellyfin from its own subdomain (e.g jellyfin.example.org) so it would be good to test this.

general/networking/nginx.md
```diff
@@ -10,65 +10,87 @@ title: Nginx Reverse Proxy
...
-#        proxy_set_header Connection "upgrade";
-#        proxy_set_header Host $host;
-#        proxy_set_header X-Real-IP $remote_addr;
-#        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-#        proxy_set_header X-Forwarded-Proto $scheme;
-#        proxy_set_header X-Forwarded-Protocol $scheme;
-#        proxy_set_header X-Forwarded-Host $http_host;
-#    }
-#}
+    # listen 443 ssl http2;
```
mcarlton00: Haven't had a chance to test yet (afk until tomorrow, sorry) but I personally don't like this change to the comments.  If we're going to have large blocks commented out, I'd prefer them all to be in a consistent line so they can easily be removed in an editor.  having them staggered and indented means a lot of manual work.  Plus it's commented out here, but uncommented out at L54.  The mix and match doesn't make sense.  

Previously in chat we'd discussed removing the commented out lines entirely and just having 2 blocks, one for 80 and one for 443 without doing the "uncomment this for SSL" thing.  I don't remember the result of that discussion though.  @Artiume, do you?

general/networking/nginx.md
```diff
@@ -10,65 +10,87 @@ title: Nginx Reverse Proxy
...
+    # listen [::]:443 ssl http2;
+    server_name DOMAIN_NAME;
+
+    # use a variable to store the upstream proxy
+    # in this example we are using a hostname which is resolved via DNS
+    # (if you aren't using DNS remove the resolver line and change the variable to point to an IP address e.g `set $jellyfin 127.0.0.1`)
+    set $jellyfin jellyfin;
+    resolver 127.0.0.1 valid=30;
```
mcarlton00: I'll be honest, I've never actually seen the benefit of doing this over just `proxy_pass http://$ip_address` or `proxy_pass http://$hostname` unless you're using nginx as a load balancer.  Just seems like more lines of code to accomplish the same thing.  I'm not opposed, I just want to understand why this is preferred.

general/networking/nginx.md
Artiume: With finding a lot of people on reddit exposing their servers, it might be better to give warnings of using http, provide a copypasta config and hope they use it to learn only. I'd like to make it a narrative of basic, advanced, more advanced.

general/networking/nginx.md
aidanharris: This is just my personal preference but I prefer using variables like `set $jellyfin IP_ADDRESS_OR_HOSTNAME;` because it makes it easier to update in the future (you don't have to change the IP address or hostname in multiple places just at the top in one place). It also means this config can mostly be used as is just by updating the variable to point to the right IP or hostname and removing the resolver line if that's not necessary (it's necessary in my case because I'm using an lxc container without a static IP so everything is dynamic).

aidanharris:
I've added a warning on using HTTP. I wasn't sure how to word this so am open to suggestions if you think this should be worded differently.

general/networking/nginx.md
```diff
@@ -7,76 +7,106 @@ title: Nginx Reverse Proxy
...
+
+    # Content Security Policy
+    # See: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
+    # Enforces https content and restricts JS/CSS to origin
+    # External Javascript (such as cast_sender.js for Chromecast) must be whitelisted.
+    #add_header Content-Security-Policy "default-src https: data: blob:; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' https://www.gstatic.com/cv/js/sender/v1/cast_sender.js; worker-src 'self' blob:; connect-src 'self'; object-src 'none'; frame-ancestors 'self'";
+
+    location / {
+        # Proxy main Jellyfin traffic
+        proxy_pass http://$jellyfin:8096;
```
mcarlton00: For reasons I don't understand yet, having a trailing slash here breaks things.  And given my previous debates with @Artiume we want the trailing slashes everywhere possible.  What doesn't make sense is that `http://192.168.0.10:8096/;` works fine, but `http://$jellyfin:8096/;` doesn't

general/networking/nginx.md
```diff
@@ -7,76 +7,106 @@ title: Nginx Reverse Proxy
...
+        proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_set_header X-Forwarded-Protocol $scheme;
+        proxy_set_header X-Forwarded-Host $http_host;
+
+        # Disable buffering when the nginx proxy gets very resource heavy upon streaming
+        proxy_buffering off;
+    }
+
+    # location block for /web - This is purely for aesthetics so /web/#!/ works instead of having to go to /web/index.html/#!/
+    location ~ ^/web/$ {
```
mcarlton00: I don't understand the benefit of this.  Why would somebody type `/web/#!` instead of just the root domain, or even just `/web` (which doesn't work)?

Artiume: https://github.com/jellyfin/jellyfin/issues/2593

I think it could be for this

general/networking/nginx.md
Artiume: Perhaps we shouldn't set $jellyfin? @aidanharris any luck on getting this to work?

general/networking/nginx.md
```diff
@@ -7,76 +7,106 @@ title: Nginx Reverse Proxy
...
+    }
+}
 ```
 
-## Nginx with Subpath
+## Nginx with Subpath (example.org/jellyfin)
 
 When connecting to server from a client application, enter `http(s)://DOMAIN_NAME/jellyfin` in the address field.
 
 Set the base URL field in the Jellyfin server.  This can be done by navigating to the Admin Dashboard -> Networking -> Base URL in the web client.  Fill in this box with `/jellyfin` and click Save.  The server will need to be restarted before this change takes effect.
```
Artiume:
```suggestion
Set the [base URL](https://jellyfin.org/docs/general/networking/index.html#base-url) field in the Jellyfin server.  This can be done by navigating to the Admin Dashboard -> Networking -> Base URL in the web client.  Fill in this box with `/jellyfin` and click Save.  The server will need to be restarted before this change takes effect.
```

anthonylavado:
@aidanharris Hi Aidan, we have an important licensing update and this will affect your contribution in this Pull Request. Please review this issue and let us know: https://github.com/jellyfin/jellyfin-docs/issues/308

Thank you.

general/networking/nginx.md
EraYaN: Please use the actual xref links, not direct links to the documentation. This way we can move it around if needed.

Artiume:
I'd like to include this
https://forum.jellyfin.org/t/nginx-useful-tips-and-tricks/1978

general/networking/nginx.md
fhriley: You should do this with an [upstream](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#upstream):

```
upstream jellyfin {
	server jellyfin:8096;
}
server {
	...
	location / {
		proxy_pass http://jellfyin;
	}
}
```

Using variables is [discouraged](http://nginx.org/en/docs/faq/variables_in_config.html):

> Using variables to store static strings is also a bad idea.

Artiume:
https://www.reddit.com/r/jellyfin/comments/glcbak/thank_nginx/fqyy01p
```
#Must be in HTTP block
proxy_cache_path  /home/cache/web 	levels=1:2    keys_zone=cWEB:50m	inactive=90d	max_size=35000m;
map $request_uri $h264Level { ~(h264-level=)(.+?)& $2; }
map $request_uri $h264Profile { ~(h264-profile=)(.+?)& $2; }

#set in Server block
proxy_cache            cWEB;
proxy_cache_valid      200 301 302 30d;
proxy_ignore_headers Expires Cache-Control Set-Cookie X-Accel-Expires;
proxy_cache_use_stale  error timeout invalid_header updating http_500 http_502 http_503 http_504;
proxy_connect_timeout 10s;
proxy_http_version 1.1;
proxy_set_header Connection "";

location /videos/
	{
	proxy_pass           http://myJF-IP:8096;
	proxy_cache_key "mydomain.com$uri?MediaSourceId=$arg_MediaSourceId&VideoCodec=$arg_VideoCodec&AudioCodec=$arg_AudioCodec&AudioStreamIndex=$arg_AudioStreamIndex&VideoBitrate=$arg_VideoBitrate&AudioBitrate=$arg_AudioBitrate&SubtitleMethod=$arg_SubtitleMethod&TranscodingMaxAudioChannels=$arg_TranscodingMaxAudioChannels&RequireAvc=$arg_RequireAvc&SegmentContainer=$arg_SegmentContainer&MinSegments=$arg_MinSegments&BreakOnNonKeyFrames=$arg_BreakOnNonKeyFrames&h264-profile=$h264Profile&h264-level=$h264Level&TranscodeReasons=$arg_TranscodeReasons";
		proxy_cache_valid    200 301 302 30d;
		}
```

dkanada:
It won't merge the changes with develop so if it failed before it will still fail now.

Artiume:
I'm actually not sure what changes with develop you mean. I also might remake the PR since the author is MIA and hasn't signed the license

PR #274, opened 2020-03-12
add subtitle extraction tip

Artiume:
untested


dkanada:
Here is some good [documentation(https://support.gitkraken.com/working-with-commits/squash) about squashing for all your branches here.

PR #276, opened 2020-03-13
Rework kodi addon instructions

mcarlton00:
Giving the kodi page some much needed love.  Reorganizing, giving much more detail where needed, and hopefully more user friendly

### DIFF
--- a/general/clients/installing-kodi.md
+++ b/general/clients/installing-kodi.md
@@ -59,7 +59,7 @@ The installation method for the repository varies depending on what kind of devi
         * Note that if you have a baseurl set, you should append that value to the end of the host field.
             * Host: `192.168.0.10:8096/jellyfin`
     * Select user account and input password, or select "Manual Login" and fill in your user infomation
-3. Once you're succesfully authenticated with the server, you'll be asked about your preferences for this device. Among these options is the choice of Add-on vs Native modes.
+3. Once you're succesfully authenticated with the server, you'll be asked about which mode you'd like to use, Add-on vs Native, which are outlined below.
 
 #### Add-on Mode
 
--- a/general/clients/installing-kodi.md
+++ b/general/clients/installing-kodi.md
@@ -1,61 +1,101 @@
 ---
 uid: clients-kodi
-title: Kodi
+title: Installing Kodi Add-On
 ---
 
-# Kodi
+# Installing Kodi Add-On
+
+> [!NOTE]
+> It's highly recommended to install the `Kodi Sync Queue` plugin into the Jellyfin server as well.
+> This will keep your media libraries up to date without waiting for a periodic re-sync from Kodi.
+
 
 ## Installation Process
 
-1. Download the repository installer [here](https://repo.jellyfin.org/releases/client/kodi/repository.jellyfin.kodi.zip).
+### Install Add-On Repository
+
+The recommended install method is to use the official Jellyfin add-on repository.  This allows for easy install of the Jellyfin for Kodi add-on, as well as automatically keeping the add-on up to date with the latest version.  Any other Jellyfin related add-ons that may be built in the future will also be available in this repository.
+
+#### General Use Devices (PCs and Tablets)
+
+1. Download the repository installer found [here](https://repo.jellyfin.org/releases/client/kodi/repository.jellyfin.kodi.zip).
     * It will be saved as `repository.jellyfin.kodi.zip`
 2. Install the Jellyfin repository.
-    * Open Kodi and navigate to "Add-On Browser"
+    * Open Kodi, go to the settings menu,  and navigate to "Add-On Browser"
     * Select "Install from Zip File"
-        * If prompted, enter settings and enable "Unknown Sources", then go back
+        * If prompted, enter settings and enable "Unknown Sources", then go back to the Add-On Browser
     * Select the newly downloaded file and it will be installed
-3. Install Jellyfin for Kodi.
+
+#### "Embedded" Devices (Android TV, Firestick, and other TV Boxes)
+
+1. Open Kodi, go to the settings menu, and navigate to "File manager"
+    * Select "Add source"
+    * In the text box, enter `https://repo.jellyfin.org/releases/client/kodi/`
+    * Enter a name for the data source, such as "Jellyfin Repo" and select Ok
+2. From the settings menu, navigate to "Add-On Browser"
+    * Select "Install from Zip File"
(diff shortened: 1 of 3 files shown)


general/clients/installing-kodi.md
```diff
@@ -1,61 +1,101 @@
...
+        * Note that if you have a baseurl set, you should append that value to the end of the host field.
+            * Host: `192.168.0.10:8096/jellyfin`
+    * Select user account and input password, or select "Manual Login" and fill in your user infomation
+3. Once you're succesfully authenticated with the server, you'll be asked about your preferences for this device. Among these options is the choice of Add-on vs Native modes.
 
-> [!NOTE]
-> It's recommended to install the `Kodi Sync Queue` plugin into the Jellyfin server as well.
+#### Add-on Mode
+
+Add-on mode uses the Jellyfin server to translate media files from the filesystem to Kodi.  This is the default setting for the add-on, and is sufficient for most use cases.  It will work both on the local network and over the Internet through a reverse proxy or VPN connection.  Providing network speed is sufficient, Kodi will direct play nearly all files and put little overhead on the Jellyfin server.  Exceptions to this rule are files with 7.1 audio tracks and some 4k content, which will be transcoded by the server.
```
Artiume: Does 5.1 transcode or is Kodi able to properly handle it?

general/clients/installing-kodi.md
```diff
@@ -1,61 +1,101 @@
...
+    * Install `repository.jellyfin.kodi.zip`
+
+### Install Jellyfin for Kodi Add-On
+
+1. Install Jellyfin for Kodi.
     * From within Kodi, navigate back to "Add-On Browser"
     * Select "Install from Repository"
-    * Choose "Kodi Jellyfin Addons"
-    * "Video Add-ons"
+    * Choose "Kodi Jellyfin Addons", followed by "Video Add-ons"
```
Artiume: add-ons or addons? We have a mixture of both

dkanada: addons imo

Artiume: i think add-on would be correct, but i'm just aiming for consistency

https://www.merriam-webster.com/dictionary/add-on

mcarlton00: All of Kodi's menus say "add-ons", but our repository says "addons".  Which I don't want to change now because that'll be annoying for users

general/clients/installing-kodi.md
mcarlton00: Kodi is able to play just about everything.  There's some incompatibility in the server where 7.1 audio will get transcoded down to 5.1

Artiume: Can you provide me some transcode logs? It might just be a simple profile issue

general/clients/installing-kodi.md
Artiume: I doubt they'd even notice either way lol

general/clients/installing-kodi.md
mcarlton00: Nope, nope I can not.  I don't have any.  That's what's been reported to me several times by other users as the reason they needed to have Native mode working

general/clients/installing-kodi.md
mcarlton00: Thinking more, the repository could be renamed, but it's not in git so would have to be manual.  It would just be changed for all future users, existing ones wouldn't change.

Need to unzip this file, change the label inside the xml file, and zip it back up. https://repo.jellyfin.org/releases/client/kodi/repository.jellyfin.kodi.zip.

I don't have shell access to those servers, so if somebody wants to do it then have at it

general/clients/installing-kodi.md
```diff
@@ -1,61 +1,101 @@
 ---
 uid: clients-kodi
-title: Kodi
+title: Installing Kodi Add-On
```
Artiume: Do we want have Installing in the title? There's more here than just installing it

mcarlton00: When it's installed it automatically launches into the configuration page.  And I was mostly basing this off what we did for the Mopidy page.

The URL is already "installing-kodi.html", so I was making that match. https://jellyfin.org/docs/general/clients/installing-kodi.html

general/clients/installing-kodi.md
```diff
@@ -1,61 +1,101 @@
...
+    * Select "Add source"
+    * In the text box, enter `https://repo.jellyfin.org/releases/client/kodi/`
+    * Enter a name for the data source, such as "Jellyfin Repo" and select Ok
+2. From the settings menu, navigate to "Add-On Browser"
+    * Select "Install from Zip File"
+        * If prompted, enter settings and enable "Unknown Sources", then go back to the Add-On Browser
+    * Select the data source you just added
+    * Install `repository.jellyfin.kodi.zip`
+
+### Install Jellyfin for Kodi Add-On
```
Artiume: Can we bump this section up to LOC28? Trying to install via this and I was confused by having to skip the Embedded devices section. 

Better yet, maybe add a step to `See Install Jellyfin for Kodi Add-On`

mcarlton00: That's why I used different sized headings, though that could be made clearer with a line "there's two options to install the repository".  I'd like to avoid a situation where you have to jump up and down the page like in the current doc to find the info you need

Artiume: i'm fairly unfamiliar with kodi so i'm fairly dependent on the docs. once I had installed the zip file, i went back to the docs trying to find the next section and `Install Jellyfin for Kodi Add-On` didnt really seem like it's the next section to perform so i became quite confused as to what to do next lol.

general/clients/installing-kodi.md
dkanada: I'm guessing the server will always max out at 5.1 for an arbitrary reason, you could probably find and remove it with no side effects.

general/clients/installing-kodi.md
Artiume: We could rename the file and the title to just Kodi

mcarlton00: If that's preferred.  We should also rename the Mopidy page as well in that case

general/clients/installing-kodi.md
mcarlton00: I added more links on next steps

general/clients/installing-kodi.md
dkanada: I agree that a general name makes more sense if we include more clients.

Artiume:
Any last thoughts @dkanada

PR #284, opened 2020-03-27
More mopidy changes

mcarlton00:
* Moves mopidy page from `installing-mopidy` to just `mopidy`
* Make Github link to point to jellyfin instead of my fork
* Change to `extension` instead of `plugin` (the term they use upstream)


mcarlton00:
My git history is all kinds of broken here, will redo

PR #293, opened 2020-04-05
Add Links to top of document to aid navigation

rexbron:
To ease navigation of the installation document, include links to the subsections at the head of the document.


Artiume:
Looks good to me. One of my goals is to split these out to their own pages and leave this page has general guidelines. Links to troubleshooting and so on.

PR #300, opened 2020-04-11
Do not specify an explicit NuGet source in the CI build

mark-monteiro:
This was added to try to fix an issue with Azure, but does not appear to be necessary


dkanada:
Can this be closed now?

mark-monteiro:
> 
> 
> Can this be closed now?

I think we still want to remove the `--source` argument. It won't be needed any more if https://github.com/jellyfin/jellyfin/pull/2834 gets merged

PR #301, opened 2020-04-11
Update Troubleshooting for inotify 

Artiume:
moving RTM to troubleshooting section. Every time I go looking for it, I naturally go to t/s first.


Artiume:
what's the DocFX turn around time? This was merged two days ago but the docs arent updating

https://jellyfin.org/docs/general/administration/troubleshooting.html

PR #304, opened 2020-04-15
Add menu for style guides and start Javascript guide

MrTimscampi:
Opening this early to allow discussion by @jellyfin/web members.

This adds a new menu "Style Guides" to the Contributing section, with a sub-entry Javascript.

For now, the style guide intends to be a mix of Google, Standard and Airbnb, the three most popular guides, taking and leaving what we find to be good practice.

### DIFF
--- a/general/style-guides/index.md
+++ b/general/style-guides/index.md
@@ -1,0 +1,10 @@
+---
+uid: style-guides
+title: Style Guides
+---
+
+# Style Guides
+
+This section documents the code style used for the different languages used by Jellyfin.
+
+If the language you're looking for doesn't have a style guide yet, respect the style of the surrounding code in the files you are editing.
--- a/general/style-guides/javascript.md
+++ b/general/style-guides/javascript.md
@@ -1,17 +1,25 @@
 ---
 uid: style-guides-javascript
-title: Javascript
+title: JavaScript
 ---
 
-# Javascript
+# JavaScript
 
-## Filenames
+## Tools
 
-Filenames must be camel case and may not include underscores (`_`) or dashes (`-`). The filename's extensions must be `.js`.
+Projects generally provide EditorConfig and ESLint configurations to help you adhere to this style guide.
+
+If one is not available or if you wish to start a new project, feel free to copy the existing files from one of the existing JavaScript projects.
 
 ## File Structure
 
-All files must abide by the following general structure, with JSDoc comments being optional but prefered:
+### Filenames
+
+Filenames must be camel case and may not include underscores (`_`) or dashes (`-`). The filename's extensions must be `.js`.
+
+### General Structure
+
+All files must abide by the following general structure, with JSDoc comments being optional but preferred:
 
 ```javascript
 /**
@@ -24,6 +32,8 @@ import module from 'dependency';
 import { myFunction, myClass } from 'dependency/submodule';
 import 'otherDependency';
 
+export const name = 'Some name';
+
 /**
  * Defines a non-exported function, accessible only from this module.
  *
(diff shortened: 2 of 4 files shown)


general/style-guides/index.md
```diff
@@ -0,0 +1,10 @@
+---
+uid: style-guides
+title: Style Guides
+---
+
+# Style Guides
+
+This section documents the code style used for the different languages used by Jellyfin.
```
nielsvanvelzen, line 8: Might be worth noting that we provide [.editorconfig] files to enforce _some_ of the guidelines (like whitespaces & line endings).

[.editorconfig]: https://editorconfig.org/

MrTimscampi: I do intend to mention that we have EditorConfig and ESLint files to help people stick to whatever will be in there :)

general/style-guides/javascript.md
```diff
@@ -0,0 +1,55 @@
...
+ * Defines a non-exported function, accessible only from this module.
+ *
+ * @param {Object} argument - The argument to pass to the function.
+ * @returns {Int|null} The resulting object from the function.
+ */
+function privateFunction (argument) {
+    // Code ommitted
+}
+
+export publicFunction (argument) {
```
dmitrylyzo, line 37: function?

general/style-guides/javascript.md
```diff
@@ -0,0 +1,55 @@
...
+    // Code ommitted
+}
+
+export default { publicFunction }
```
dmitrylyzo, line 41: maybe multi line example?

general/style-guides/javascript.md
```diff
@@ -0,0 +1,190 @@
...
+### Braces
+
+Braces are required for all control structures, even if the body contains only a single line.
+
+They should respect the following rules:
+
+* No line break before the opening brace
+* Line break after the opening brace
+* Line break before the closing brace
+* Line break after the closing brace, except when the brace is followed by `else`, `catch`, `while`, or a comma, semicolon, or right-parenthesis.
```
dkanada, line 81: none of the other bullets have periods

general/style-guides/javascript.md
```diff
@@ -0,0 +1,190 @@
...
+    publicFunction
+}
+
+```
+
+### Imports
+
+#### Naming
+
+If the imported module doesn't respect the file naming conventions, convert its name to camel case when importing its default export, like such:
```
dkanada, line 62: I'm really not a fan of ending sentences with colons.

general/style-guides/javascript.md
```diff
@@ -0,0 +1,190 @@
...
+### Variables
+
+Variable declarations must use `let` or `const`. Use `const` by default, unless the variable needs to be reassigned. Avoid the use of the `var` keyword.
+
+Declare variable as close to their point where they are first used as possible, in order to their scope.
+
+> [!NOTE]
+> Avoid declaring multiple variables in one declaration.
+
+
```
dkanada, line 121:
```suggestion

```

general/style-guides/javascript.md
```diff
@@ -0,0 +1,190 @@
...
+> [!IMPORTANT]
+> Do not use trailing commas when declaring arrays.
+
+> [!INFO]
+> Do not declare objects with `new Object()`. While it doesn't share the issues of the Array constructor, consistency is important.
+
+Do not mix quoted and unquoted keys in a single object literal.
+
+
```
dkanada, line 173:
```suggestion

```

general/style-guides/javascript.md
MrTimscampi: Currently, it is required at least in Web, since adding colons to everything recently actually fixed a few bugs...

Not using semi-colons is also error-prone since browsers WILL insert semicolons internally while evaluating (And so will transpilers and minifiers). These parsers have rules for how they parse statements which, if you don't know them, can cause issues.

For example, this:

```javascript
a = b + c 
(d + e).print() 
```

Would turn into this:

```javascript
a = b + c(d + e).print();
```

These issues can be hard to debug and, imo, in the interest of avoiding risk and ensuring we have less bugs introduced by accident, we should enforce semicolon usage.

general/style-guides/javascript.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+In cases where an object only has one element, it an be declared on one line, as follow:
+
+```javascript
+const myObject = { a: 1 };
+```
+
+> [!IMPORTANT]
+> Do not use trailing commas when declaring arrays.
```
YouKnowBlom, line 166: Shouldn't this say "when declaring objects."?

PR #309, opened 2020-04-18
Remove release step as a requirement

dkanada:
@EraYaN this is some of the required work. We'll need to either port the commit script to PowerShell or the build job to Linux, I vote for the latter option. We also need to add the jellyfin-bot SSH key to Azure so it can be used in the script, but those are the only things left.

### DIFF
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -72,17 +72,32 @@ jobs:
         artifactName: 'jellyfin-docs-$(Build.BuildId)'
         publishLocation: 'Container'
 
+  - job: Publish
+    displayName: 'Publish'
+
+    dependsOn: Build
+    condition: contains(variables['Build.SourceBranch'], 'master')
+
+    pool:
+      vmImage: ubuntu-latest
+
+    steps:
+    - checkout: none
+
+    - task: DownloadPipelineArtifact@2
+      displayName: 'Download Artifact'
+      inputs:
+        source: 'current'
+        artifact: 'jellyfin-docs-$(Build.BuildId)'
+        path: '$(Build.SourcesDirectory)'
+        runVersion: 'latest'
+
     - task: DownloadSecureFile@1
       displayName: 'Download GitHub Key'
-      condition: contains(variables['Build.SourceBranch'], 'master')
       name: 'Bot'
       inputs:
         secureFile: 'bot'
 
-    - task: CmdLine@2
+    - script: 'bash .ci/publish-to-gh-pages.sh'
       displayName: 'Update Website'
-      condition: contains(variables['Build.SourceBranch'], 'master')
-      inputs:
-        targetType: 'filePath'
-        filePath: '.ci/publish-to-gh-pages.sh'
-        workingDirectory: '$(Build.ArtifactStagingDirectory)'
+      workingDirectory: '$(Build.SourcesDirectory)'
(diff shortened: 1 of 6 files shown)


.ci/azure-pipelines.yml
```diff
@@ -1,76 +1,74 @@
...
-        archiveFile: '$(Build.ArtifactStagingDirectory)/docs-$(Build.BuildId).tar.gz'
-        replaceExistingArchive: true
+        archiveType: 'zip'
+        archiveFile: '$(Build.ArtifactStagingDirectory)/jellyfin-docs-$(Build.BuildId).zip'
 
-    - task: GitHubRelease@0
-      condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'), in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI', 'BuildCompletion'))
-      displayName: "Create GitHub Release"
+    - task: PublishBuildArtifacts@1
+      displayName: 'Publish Docs'
```
EraYaN, line 70: This needs the same condition as the previous task. Otherwise there is nothing to publish.

.ci/publish-to-gh-pages.sh
```diff
@@ -1,18 +1,21 @@
 #!/bin/bash
 
-# Download the latest release
-curl -s https://api.github.com/repos/jellyfin/jellyfin-docs/releases/latest | grep "browser_download_url.*docs-.*\.tar\.gz" | cut -d : -f 2,3 | tr -d \" | wget -O /tmp/docs.tar.gz -qi -
```
EraYaN, line 4: you need to replace this with some thing like:

```sh
azure_build_id="$( az pipelines runs list --branch release-10.5.z --pipeline-ids 1 --reason batchedCI --result succeeded --status completed --query-order QueueTimeDesc --top 1 | jq '.[].id' )"
az pipelines runs artifact download --artifact-name 'Jellyfin Server Setup' --path ${installer_tmpdir} --run-id ${azure_build_id}
```

dkanada: Something like that or can I just use that exact command? It looks fine to me.

.ci/publish-to-gh-pages.sh
```diff
@@ -1,18 +1,24 @@
...
+# move to git directory
+cd jellyfin.github.io
+git add docs
 
-git add docs/
-git commit -m "CI Documentation update"
-git push origin
\ No newline at end of file
+# commit new changes
+git -c "user.name=jellyfin-bot" -c "user.email=team@jellyfin.org" commit -m 'Azure Update ${BUILD_BUILDID}'
```
mark-monteiro, line 17: Perhaps the commit message should specify that this is an update for the docs section of the site

dkanada: done

.ci/publish-to-gh-pages.sh
EraYaN: It need different arguments, but yes it's the exact command that the build server uses to pull in the windows setup artifact.

dkanada:
@EraYaN I just added a step to download the artifact in the pipeline rather than with a script, this should be ready for another review now.

EraYaN:
One last thing, the current publish condition should contain a check for "BatchedCI", "BuildCompletion" and "IndividualCI" because the current condition becomes: `contains('refs/pull/309/merge', 'master')` might at some point be exploited I feel by some PR or someone starting a build manually.

dkanada:
Ah, I should add the protection in another commit.

PR #311, opened 2020-04-18
Use Ubuntu image for CI instead of Windows

mark-monteiro:


EraYaN:
Is there a reason to switch to Linux from Windows now? Or is that more for better on-boarding of other parts of the project?

mark-monteiro:
I guess it's more consistent with the CI for the other projects, but I'm not really sure. @dkanada requested it

dkanada:
I suggested it because most of us mainly use Linux, and 90% of our pipelines use Linux as well so it helps standardize things.

Issue #273, opened 2020-03-12
HWA with Nvidia GPUs + docker


D34DC3N73R:
I see a couple of issues with the documentation in regards to docker + hwa + nvidia gpus.

1.  Nvidia GPUs are supported when using docker-compose.
Prerequisites: nvidia drivers, nvidia docker 2 package, edited `/etc/docker/daemon.json` to make nvidia the default runtime in docker-compose v3+ (as you have outlined). Simply add the nvidia environment variables to the docker-compose jellyfin service.
```
    environment:
      - NVIDIA_VISIBLE_DEVICES=all
      - NVIDIA_DRIVER_CAPABILITIES=all
      ...
```
This should be compatible with any glibc distribution (which is required for nvidia drivers and nvidia docker 2).

2. When using docker v19.03+ (docker run), there is no reason to edit `/etc/docker/daemon.json` as gpu options have been integrated into the default runtime. The nvidia docker 2 package is no longer needed, but [nvidia container toolkit](https://github.com/NVIDIA/nvidia-docker) is required as well as nvidia drivers. You no longer need the specific nvidia environment variables or switching to the nvidia runtime. Just use  `--gpus all`
```
docker run -d \
--name=jellyfin \
--gpus all \
-p 8096:8096 \
-p 8920:8920 \
-v /config:/config \
-v /media:/media \
-v /cache:/cache \
--restart unless-stopped \
jellyfin/jellyfin
```
Again, this should be compatible with any glibc distribution (which is required for nvidia drivers and nvidia container toolkit).

Artiume:
Would you be willing to get a working config using `user: 1000:1000` and `group_add: videogroup` I know of group 44 and 107 and had issues getting both groups added to the container. It's been hard getting the permissions right

D34DC3N73R:
When using nvidia container toolkit / nvidia docker 2, you do not need to specify or add an alternate video group. This is from my docker-compose.yml. PUID/PGID .env is set to admin:admin (1000). 
```
  jellyfin:
    image: linuxserver/jellyfin
    container_name: jellyfin
    environment:
      - PUID=$PUID
      - PGID=$PGID
      - TZ=$TZ
      - NVIDIA_VISIBLE_DEVICES=all
      - NVIDIA_DRIVER_CAPABILITIES=all
    volumes:
      - $HOME/.config/jellyfin:/config
      - /mnt/media:/media
      - /dev/shm:/transcode
    ports:
      - 8095:8096
[...]

PR #312, opened 2020-04-18
Improved Docker installation Documentation

ThijsvL:
Improved the Docker Documentation to be a little more "Noob" friendly, and added a way to get started with docker by adding the installation steps.

### DIFF
--- a/general/administration/installing.md
+++ b/general/administration/installing.md
@@ -65,15 +65,11 @@ curl -fsSL https://get.docker.com -o get-docker.sh
 sh get-docker.sh
 ```
 
-And to install Docker Compose you will need to run the following commands:
+And to install Docker Compose you will need to run the following command:
 
 ```sh
-sudo curl -L "https://github.com/docker/compose/releases/download/1.25.5/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
-sudo chmod +x /usr/local/bin/docker-compose
+apt-get install docker-compose
 ```
-
-> [!NOTE]
-> To install a different version of Compose, substitute 1.25.5 with the version of Compose you want to use.
 
 Now test the installation by running the following command in your terminal:
 
--- a/general/administration/installing.md
+++ b/general/administration/installing.md
@@ -14,7 +14,7 @@ The Jellyfin project and its contributors offer a number of pre-built binary pac
       - [Installing Docker](#installing-docker)
       - [Preparing the Folders](#preparing-the-folders)
       - [Running the Container](#running-the-container)
-    - [Docker Hub image maintained by LinuxServer.io](#docker-hub-image-maintained-by-linuxserverio)
+      - [Docker Hub image maintained by LinuxServer.io](#docker-hub-image-maintained-by-linuxserverio)
     - [Unraid Docker](#unraid-docker)
     - [Kubernetes](#kubernetes)
   - [Windows (x86/x64)](#windows-x86x64)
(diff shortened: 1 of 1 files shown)


Artiume:
Any reason to remove the lsio link?

ThijsvL:
> Any reason to remove the lsio link?

Yes, as the Official Docker Header did not have a link but a button, for consistancy I thought it would be nice to have the Ls.io header not be a link but only the button to have a link.

ThijsvL:
> Any reason that you install the docker-compose from the git release instead of apt, i know its a bit older than latest but just as good and is more noob friendly.

I'll have to agree with you, apt-get is easier, my thought was that it was friendlier for people using different systems then Ubuntu + newer release, but you can't go wrong with apt-get

nekocentral:
If i provide the commands and a basic explantion for the docker nvdia part we can mabye also make a noob friendly addition (just did it my self) as i can see your better at writing than me

nekocentral:
# Add the package repositories
distribution=$(. /etc/os-release;echo $ID$VERSION_ID)
curl -s -L https://nvidia.github.io/nvidia-docker/gpgkey | sudo apt-key add -
curl -s -L https://nvidia.github.io/nvidia-docker/$distribution/nvidia-docker.list | sudo tee /etc/apt/sources.list.d/nvidia-docker.list

#Install Nvidia
sudo apt-get update && sudo apt-get install -y nvidia-container-toolkit nvidia-docker2
sudo systemctl restart docker

#Docker compose
runtime: nvidia
    environment:
      - NVIDIA_VISIBLE_DEVICES=all

also make the docker-compose.yml atleast version 2.3

Artiume:
What about the idea of dropping the the compose version to 2 so that group_add can be added. It's needed to do hwa with our image.

nekocentral:
Just did some testing

nvidia-docker2 is not needed anymore and so does it seem for "runtime: nvidia"

nekocentral:
> What about the idea of dropping the the compose version to 2 so that group_add can be added. It's needed to do hwa with our image.
> 
> Edit :
> Didn't see your comment in time. X)
> Adding this #273
I would personally use 2.3 so you can use the same docker file with nvenc encoding



> #Add the package repositories
> distribution=$(. /etc/os-release;echo $ID$VERSION_ID)
> curl -s -L https://nvidia.github.io/nvidia-docker/gpgkey | sudo apt-key add -
> curl -s -L https://nvidia.github.io/nvidia-docker/$distribution/nvidia-docker.list | sudo tee /etc/apt/sources.list.d/nvidia-docker.list
> 
> #Install Nvidia
> sudo apt-get update && sudo apt-get install -y nvidia-container-toolkit nvidia-docker2
> sudo systemctl restart docker
> 
> #Docker compose
> runtime: nvidia
> environment:
> - NVIDIA_VISIBLE_DEVICES=all
> 
> also make the docker-compose.yml atleast version 2.3

I used this docker-compose.yml for the docker-nvidia part
```
---
version: "2.3"
services:
  jellyfin-test:
    image: jellyfin/jellyfin
    container_name: jellyfin-test
    user: 1000:1000
    runtime: nvidia
    environment:
      - TZ=$TZ
      - NVIDIA_VISIBLE_DEVICES=all
    volumes:
      - /config/test/config:/config
      - /mnt/tv:/media
      - /tmp:/transcode
    ports:
      - 8096:8096
```

ThijsvL:
On it!

Artiume:
I like the updates. I'm ready to merge it. I think the next step should be to break apart all of the install docs to an index and separate pages for each system. Sort of like I did for the network section.

PR #314, opened 2020-04-19
Include list of supported browsers

Artiume:


general/clients/index.md
```diff
@@ -12,6 +12,25 @@ Clients connect your devices to your Jellyfin server and let you view your conte
...
+ - last 2 ChromeAndroid versions
+ - last 2 Safari versions
+ - last 2 iOS versions
+ - last 2 Edge versions
+ - Chrome 27
+ - Chrome 38
+ - Chrome 47
+ - Chrome 53
+ - Chrome 56
+ - Chrome 63
```
MrTimscampi, line 30: All these specific Chrome versions are basically for Tizen and WebOS, not really for desktops (though it should still work)

general/clients/index.md
```diff
@@ -12,6 +12,25 @@ Clients connect your devices to your Jellyfin server and let you view your conte
...
 
+## Browsers
+
+Our goal is to provide support for these browsers. 
+
+ - last 2 Firefox versions
+ - last 2 Chrome versions
+ - last 2 ChromeAndroid versions
+ - last 2 Safari versions
+ - last 2 iOS versions
```
MrTimscampi, line 23: Should probably be grouped with the Safari line and changed to "last 2 versions of Safari of MacOS and iOS"

general/clients/index.md
```diff
@@ -12,6 +12,25 @@ Clients connect your devices to your Jellyfin server and let you view your conte
 
 Do you have a client that interfaces with Jellyfin and want to see it listed here? Please [submit a pull request](https://github.com/jellyfin/jellyfin-docs)!
 
+## Browsers
+
+Our goal is to provide support for these browsers. 
+
+ - last 2 Firefox versions
+ - last 2 Chrome versions
+ - last 2 ChromeAndroid versions
```
MrTimscampi, line 21: Should probably be changed to "last 2 Chrome for Android versions"

general/clients/index.md
Artiume: If you know which go with which, i can add them in parenthesis

general/clients/index.md
Artiume: What about older tablets? Would they need back support like chrome does?

MrTimscampi: If they're still getting iOS updates, they should be good, I think? I don't use Apple products, so I'm not fully sure what their support policy is.

Artiume: I refactored the layout a little, tell me what you think

general/clients/index.md
dkanada: We should make a note that we aren't opposed to supporting older browsers through polyfills and such, but it's not a guarantee.

Artiume: added

general/clients/index.md
```diff
@@ -12,6 +12,24 @@ Clients connect your devices to your Jellyfin server and let you view your conte
...
 Do you have a client that interfaces with Jellyfin and want to see it listed here? Please [submit a pull request](https://github.com/jellyfin/jellyfin-docs)!
 
+## Browsers
+
+Our goal is to provide support for the two most recent versions of these browsers. Older browsers may be supported, but are not guaranteed. 
+ - Firefox
+ - Chrome
+ - Chrome for Android
+ - Safari for MacOS and iOS
+ - Edge
```
MrTimscampi, line 22: For Edge, you can maybe do this:
```suggestion
 - Edge (EdgeHTML and Chromium engines)
```

general/clients/index.md
MrTimscampi: We should probably remove all the Chrome versions and move "Older browsers may be supported, but are not guaranteed." after the list, as something like this:

```markdown
Older browsers may be supported as a result of the needs of specific web-based clients, but full functionality is not guaranteed on their desktop version.
```

PR #324, opened 2020-04-25
shows: add a note about named season folders

MayeulC:
I somehow forgot to open the PR after committing.

I hit this issue, and apparently it's quite a common one, especially with SG-1,
so let's document it :)
The tip is a bit verbose, but I tried to word it to be berrter referenced when troubleshooting using a search engine :)

### DIFF
--- a/general/server/media/shows.md
+++ b/general/server/media/shows.md
@@ -23,4 +23,5 @@ The most common naming scheme for shows is categorizing the files by series and 
         Episode S02E02.mp4
 ```
 ----
-Note that the season folders *do not contain the series name*, otherwise jellyfin can in certain cases (Stargate SG-1 due to the dash, for instance) misdetect your episodes and put them all under the same season.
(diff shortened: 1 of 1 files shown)


general/server/media/shows.md
```diff
@@ -22,3 +22,5 @@ The most common naming scheme for shows is categorizing the files by series and
         Episode S02E01.mp4
         Episode S02E02.mp4
 ```
+----
+Note that the season folders *do not contain the series name*, otherwise jellyfin can in certain cases (Stargate SG-1 due to the dash, for instance) misdetect your episodes and put them all under the same season.
```
Artiume, line 26:
```suggestion
> [!NOTE]
> Season folders shouldn't contain the series name, otherwise Jellyfin can in certain cases (Stargate SG-1 due to the dash and one, for instance) misdetect your episodes and put them all under the same season.
```

Issue #318, opened 2020-04-24
Update documenation wording to be less docker specific


rigtorp:
Docker is one of many container technologies. Podman is the preferred container solution on Fedora/Redhat/Centos for desktop/workstation and kubernetes is the preferred solution for servers. Synology etc will likely move to kubernetes as the supported container solution in the future. 

Let's add podman and kubernetes instructions. See #317 for podman instructions.

Artiume:
I agree. Why not create a PR? It would help it get implemented faster :)

rigtorp:
Yes I plan to. Submitted first PR with instructions for podman: #329

PR #331, opened 2020-04-28
Update build docs to use generic container terminilogy

rigtorp:
Related to #318

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /**/bin/
 /**/obj/
 _site
+node_modules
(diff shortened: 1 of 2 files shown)


Artiume:
I've never built a container with podman/buildah. Would it be worth it adding that as well in another PR?

rigtorp:
buildah is awesome since it allows you to build images easily in build pipelines without Docker-in-Docker mess. I tried and it seems like the jellyfin Dockerfile is not compatible with buildah. Dockerfiles is not the recommended way to build images using buildah. Example of using the native buildah way https://www.redhat.com/sysadmin/building-buildah

PR #334, opened 2020-04-30
Documentation on Internet radio by using the Live TV M3U Tuner

mahriman:


general/server/media/internet-radio.md
```diff
@@ -0,0 +1,21 @@
...
+```#EXTM3U
+#EXTINF:0,Radio Freccia
+https://streamingv2.shoutcast.com/radiofreccia
+```
+
+Note that the line that starts with `#EXTINF:0,<title>` is needed for each radio URL to give it a 'channel' entry under Live TV \ Channels. Failing to add this line will cause the station to not show up under Live TV \ Channels.
+
+Next, head over to the Jellyfin administration page, go to Live TV, add new tuner device, choose M3U Tuner as Tuner type and navigate to your M3U file. Hit Save and let Jellyfin complete the Refresh Guide task (automatically started when saving a new tuner). You should now be able to play your radio station from under Live TV \ Channels.
+
+**Note**: Adding an M3U HTTP link instead of a locally created M3U file will almost certainly fail, in part because the `#EXTINF:` directive is part of the IPTV standard, which is required to name the channel for Jellyfin to list it under Live TV \ Channels. Pretty much no Internet radio will include this directive in their M3U files. Besides that, many radio stations use AJAX to dynamically update the M3U-files while listening, something that is not handled by Jellyfin.
```
Artiume, line 21: @MrTimscampi Could we support AJAX?

Artiume, line 21:
```suggestion
> [!NOTE]
> Adding an M3U HTTP link instead of a locally created M3U file will almost certainly fail, in part because the `#EXTINF:` directive is part of the IPTV standard, which is required to name the channel for Jellyfin to list it under Live TV \ Channels. Pretty much no Internet radio will include this directive in their M3U files. Besides that, many radio stations use AJAX to dynamically update the M3U-files while listening, something that is not handled by Jellyfin.
```

Artiume: I dropped in this link in the m3u 
http://yp.shoutcast.com/sbin/tunein-station.m3u?id=1796249
All of the radio stations are given the same name, but since the m3u respected the #EXTINF, it worked great :)

Artiume: Also, if you enable livetv transcoding, it causes a black video to be played with the audio. It's pretty funny, it gets encoded at 140 Mbps on my system, yet no real load.

PR #333, opened 2020-04-29
Cleanup Docker documentation

rigtorp:
It's out of scope for Jellyfin documentation to describe in detail how to install Docker or to be a tutorial on using Docker.

Additionally the tone of the previous instructions were somewhat childish.

Closes #331

PR #338, opened 2020-05-02
Add info about 3rd party image maintainer

hotio:


Artiume:
Glad to see more maintainers :). So I would look at https://github.com/jellyfin/jellyfin-docs/pull/333 to see how we're refactoring the docker section. We can add your link to your docker hub, but the detailed explanation should be cut down some. 

Checking out your docker hub, my only concern is your images, as far as I can tell, are built and pushed manually. This isn't a best practice because you could push code into your images without anyone being the wiser. I would definitely make a bigger note that you don't install drivers inside the container images so HWA support may be lacking.

hotio:
Images are not pushed manually, everything is automated and built using GitHub Actions and Drone CI....maybe you are refering to the limited DockerHub automated builds?

Artiume:
the docker section has been reworked and this needs to be rebased. I think a simple link alongside the linuxserver one should be sufficient. If there's anything unique about the image, that can be added. Is the image compatible with the jellyfin/jellyfin image?

PR #353, opened 2020-05-24
Add PublishedServerUrl parameter

neilsb:
Add PublishedServerUrl to configuration parameter list

For PR jellyfin/jellyfin#3033


neilsb:
> Does it need a `|` at the end of it? Noticing some having that.

Look like all the other do.   Added for consistency :)

PR #360, opened 2020-06-11
CSS - Set Background Images

Artiume:


general/clients/css-customization.md
```diff
@@ -69,18 +69,28 @@ This will affect the played/watched indicator. Replace the color hex with any va
...
-#### Reduce Live TV Channel Width
+### Reduce Live TV Channel Width
 
 ```css
 .channelsContainer { max-width: 8em; }
 ```
 
+### Background Image on Homepage
+
+[Additional MDN Documentation}(https://developer.mozilla.org/en-US/docs/Web/CSS/background)
```
neilsb, line 86: Should be a ]

general/clients/css-customization.md
```diff
@@ -69,18 +69,28 @@ This will affect the played/watched indicator. Replace the color hex with any va
...
 .channelsContainer { max-width: 8em; }
 ```
 
+### Background Image on Homepage
+
+[Additional MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/background)
+
+```css
+#indexPage {
+    background: url(https://i.ytimg.com/vi/avCWDDox1nE/maxresdefault.jpg);
```
shadow7412, line 90: Heh it's official now :P
Should we add `background-size: cover` if we're officialising this?

nielsvanvelzen: Might be a good idea to include the background-size too. But in that case `background` should be `background-image`

Artiume: updated

Artiume:
> 
> 
> Not that my approval really matters - but it was my code originally ;)

Na, it totally matters. Community built! Most of this CSS document has been contributed by others. The only CSS I actually know is what I've learned by maintaining the docs lol

PR #365, opened 2020-06-23
Add Traefik v2.x instructions

### LINKED ISSUE #213: Administration/Reverse Proxy: Update config for Traefik 2.0
Traefik 2.0 uses different configuration format which is not backwards compatible. /cc @PrplHaz4

michaelkrieger:
Resolves #213 
I have added a full guide to using Traefik v2.x with Jellyfin, using as much 'dynamic' configuration of Jellyfin as possible/reasonable.
I tried to make it well commented, including items not directly associated with this configuration, so it is an easy-to-adapt configuration.

I have tested this configuration as working (with the DNS-01 provider) using host networking and with DLNA, but added the necessary blocks for HTTP-01 ACME configuration as well.


michaelkrieger:
This should resolve #213

Sorry for all the updates- wanted to get the checks/lint perfect and add a but more detail in a few areas.
Should be fairly solid now.

general/networking/traefik2.md
```diff
@@ -0,0 +1,267 @@
+---
+uid: network-reverse-proxy-traefik2
+title: Traefik2
```
z3ntu, line 3: If this title is user visible I would change it to `Traefik v2` or similar

michaelkrieger: The short title is not user visible.  The title below it is, similar to as you suggest `Traefik v2.x`, and is visible to the end user.  This title is more to match the table of contents and the doc filename with a short title.

michaelkrieger: Actually in hindsight I agree.  Changing.

michaelkrieger:
Okay- I think I'm done at this point and the configuration is solid and well-tested. More detail (but well commented) was the goal so that it's easy to follow, even if it makes it longer. The configuration meets all security standards, works reliably, etc.

Open to any comments, but otherwise it should be good to go.

PR #385, opened 2020-07-21
Add native to docker migration guide

Shadowghost:
Adds documentation about how you can configure the official docker image to take in the config and data from a local install for Debuntu.

### DIFF
--- a/general/administration/installing.md
+++ b/general/administration/installing.md
@@ -124,7 +124,7 @@ Using host networking (`--net=host`) is optional but required in order to use DL
 Create a `docker-compose.yml` file with the following contents:
 
    ```yml
-   version: "2.3"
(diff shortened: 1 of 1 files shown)


general/administration/installing.md
```diff
@@ -722,3 +723,55 @@ Raw Ubuntu packages, including old versions, are available [here](https://jellyf
...
+       jellyfin/jellyfin
+   ```
+
+**Using docker-compose**
+
+   ```yml
+   version: "2"
+   services:
+   jellyfin:
+       image: bitwrk/jellyfin-rffmpeg:unstable
```
thornbill, line 762: You used a different docker image here. Probably best to use the official one in both places.

PR #391, opened 2020-07-30
Status update on Jellyfin MPP and Jellyfin desktop

cromefire:


cromefire:
Is there a reason we have separate repos for pi and windows in desktop?

dkanada:
That should probably be removed as well.

cromefire:
Removed them, put desktop under cross platform and added raspberry pi as a platform explicitly

dkanada (approved):

Issue #358, opened 2020-06-06
Outdated documentation about AMF hardware acceleration


Milokita:
Now the AMF officicaly supports linux starting from version 1.4.9
https://github.com/GPUOpen-LibrariesAndSDKs/AMF

Therefore, the doc needs to be updated, adding guide to properly install AMF for linux and configuration for docker installed version is also needed

@nyanmisaka

nyanmisaka:
The current `amdgpu-pro` 20.20 driver accidentally lost AMF related components. After it is resolved, I will try to install on Ubuntu 20.04 to write an installation guide.

PR #394, opened 2020-07-31
add guidance for AMF and QSV, update NV driver requirements

nyanmisaka:
**Issues**
https://github.com/jellyfin/jellyfin-docs/issues/358


Shadowghost:
Line 34 misses a space before `amdgpu-pro`

Another problem I see is that you can either install `intel-media-va-driver` or `intel-media-va-driver-non-free`, not both at the same time as currently written.

joshuaboniface (approved):

PR #399, opened 2020-08-05
Small change to Windows Plugin disclaimer an the there listed Paths

h1dden-da3m0n:
Since not every user is aware that the AppData/local directory is located in their users profile folder I added the %UserProfile% variable to the listed paths.

Really a minor change, but I guess this could help novice Windows users.


nielsvanvelzen:
When using the Windows Tray App the path is actually `%ProgramData%\Jellyfin\Server\plugins`. The appdata location is only when running the server directly. We might want to add that to the text.

h1dden-da3m0n:
Don't mind me I'll add this to the disclaimer then ;)

h1dden-da3m0n:
I allowed myself to rephrase the disclaimer a bit and mentioned both paths explicitly.

Hope this is fine.

PR #405, opened 2020-08-16
Simplify Traefik v2 setup

masterkoppa:
While troubleshooting an issue with my personal Traefik v2 setup, I referenced the docs to see if I was missing anything new (Thank you @michaelkrieger for putting these together!). However, I found the label setup here unnecessary and overly complex. Especially if you already have a Traefik v2 setup for other things.

Since we can't use the built in management from Traefik in host mode, I moved most of the configuration of the Jellyfin configuration to the `toml` file, instead of using labels. As a bonus, this should in theory allow folks to use a non-Docker Jellyfin installation with Traefik v2.

NOTE: Since this has a lot of placeholders and I used `yml` instead of `toml` on my own setup, I might have missed some quotes or messed something up so a second set of eyes would be appreciated.

### DIFF
--- a/general/networking/traefik2.md
+++ b/general/networking/traefik2.md
@@ -65,62 +65,6 @@ services:
       # VAAPI Devices
       - /dev/dri/renderD128:/dev/dri/renderD128
       - /dev/dri/card0:/dev/dri/card0
-    labels:
-      - "traefik.enable=true"
-      ## HTTP Router
-      #### Entry point where Jellyfin is accessible via
-      #### Change secure to https in the line below to have accessible without needing to specify a port and change the SSLHost option below
-      - "traefik.http.routers.jellyfin.entryPoints=secure"
-      #### Host or Path where Jellyfin is accessible
-      #### Remove (or change) this rule if you'd rather have Jellyfin accessible at a PathPrefix URI
-      - "traefik.http.routers.jellyfin.rule=Host(`HOST_NAME.DOMAIN_NAME`)" # OPTIONAL: && PathPrefix(`/jellyfin`)
-      #### Enable TLS with the ACME/LetsEncrypt resolver for HOSTNAME.DOMAIN_NAME
-      - "traefik.http.routers.jellyfin.tls=true"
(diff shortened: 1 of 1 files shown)


michaelkrieger:
Thanks @masterkoppa
I’d suggest, if you like this pull, _adding_ it to the documentation as an alternate configuration block instead of replacing the setup. Anyone which has or wants to set up multiple containers (torrents, pihole, VPN, NAS, etc) often wishes to keep the configuration together with the Docker container as much as possible.  To go and put the configuration _only_ in the yaml file removes the (probably) most common use case.  Another note is that it doesn’t actually -need- host networking except for certain discovery features, so to say that it can’t use host networking is incorrect.

I’ll suggest adding this good contribution instead of replacing an otherwise working/common/good configuration.  A paragraph that says “if you wish to not use labels in your configuration or if you want to use Traefik v2 without the Docker provider (ie: a non-Docker Jellyfin installation), you can replace the labels of the jellyfin service with the following code in the TOML file” and then having the code block... that I think makes more sense.

It’s a great idea to add this to the documentation. I don’t think it’s better to make this the default configuration nor better to remove the Docker provider configuration though.

masterkoppa:
@michaelkrieger 

I had debated that approach at first, but I didn't like the potential for duplication when I started it. But I didn't think of making it a potential addendum like you suggested, I'll update the PR with something like that.

joshuaboniface:
I agree with the add-rather-than-replace idea, otherwise good. CC @masterkoppa

Artiume (approved):

michaelkrieger:
Are we doing the replace based on the approval?  I don't think the new configuration is any more ideal, though adding it to the documentation as an alternate implementation is fine. Replacing a working implementation using Docker Labels for a TOML configuration for the sake of it just doesn't make sense.

Artiume:
then let's fix it or close it

dkanada:
@masterkoppa we have several comments in here requesting this method be added rather than replace the main configuration so could you make that change before this gets merged?

Issue #398, opened 2020-08-05
Offical plugin repository link should be added to the plugins section of the docs.


Darkwolf1515:
Just in case someone accidentally removes it, I don't think I've seen the link officially posted anywhere.

https://jellyfin.org/docs/general/server/plugins/index.html?q=plugin

PR #409, opened 2020-08-20
Add Default Plugin Repo link

zadi15:
Added a section for the official plugin repository link, in case someone accidentally deletes it and does not know it.
Relevant to Issue #398


mcarlton00:
Good call. We should also put a header above the repos section to more easily distinguish it from the rest of the page.  Probably `##  Repositories` would be good

zadi15:
Good idea, just done!

PR #410, opened 2020-08-24
Roku Store Logo Addition

Essasetic:
Found a Roku Store logo and thought it would look nicer than a direct link.
However it isn't direct from the Roku website (if someone can find it direct from there feel free to replace it).

### DIFF
--- a/general/clients/index.md
+++ b/general/clients/index.md
@@ -107,7 +107,7 @@ The official Jellyfin Roku app.
 
 **Links:**
 
-<a href='https://channelstore.roku.com/details/592369/jellyfin'><img alt='Get it on the Roku Store' src='https://www.scientology.tv/images/roku-badge.png' width="200"/></a>
+<a href='https://channelstore.roku.com/details/592369/jellyfin'><img alt='Get it on the Roku Store' src='https://www.roku.com/s/1597964107537/fonts/roku-logo.svg' width="200"/></a>
 
 - [GitHub](https://github.com/jellyfin/jellyfin-roku)
 
--- a/general/clients/index.md
+++ b/general/clients/index.md
@@ -107,7 +107,8 @@ The official Jellyfin Roku app.
 
 **Links:**
 
-- <a href='https://my.roku.com/add/jellyfin'><img alt='Get it on the Roku Store' src='https://www.scientology.tv/images/roku-badge.png' width="200"/></a>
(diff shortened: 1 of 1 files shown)


general/clients/index.md
```diff
@@ -107,7 +107,8 @@ The official Jellyfin Roku app.
 
 **Links:**
 
-- [Roku Channel Store](https://my.roku.com/add/jellyfin).
+<a href='https://my.roku.com/add/jellyfin'><img alt='Get it on the Roku Store' src='https://www.scientology.tv/images/roku-badge.png' width="200"/></a>
```
Maxr1998, line 110: Shouldn't this preferably use an URL from the official Roku site? Unrelated to your change, I think instead of the my.roku.com Link that requires a sign-in, a direct link to the channel store listing might be preferable: https://channelstore.roku.com/details/592369/jellyfin

Essasetic: 1. I couldn't find one like this on the Roku site. There are logos that could be used from the Roku site but it would look out of place.
2. Ah ok, that link was in the docs before so I thought to just use that link.

nielsvanvelzen: We can include the image in the repository instead.

Essasetic: Ok updated the logo. Honestly doesn't look as bad as I was anticipating but still looks kinda out of place. 

Can understand why you want it from the original repo though.

general/clients/index.md
```diff
@@ -107,7 +107,8 @@ The official Jellyfin Roku app.
 
 **Links:**
 
-- [Roku Channel Store](https://my.roku.com/add/jellyfin).
+<a href='https://channelstore.roku.com/details/592369/jellyfin'><img alt='Get it on the Roku Store' src='https://www.roku.com/s/1597964107537/fonts/roku-logo.svg' width="200"/></a>
```
nielsvanvelzen, line 110: Looks like the image isn't working anymore. I think the number part of this URL changes regularly.

Essasetic: It still works for me?

Essasetic: Oh, yeah. There is one on Wikipedia that I can link (as it claims it's under public domain). If that's ok to do.

nielsvanvelzen: Just put it in the repository please

dkanada (approved):

PR #418, opened 2020-09-22
Replaced 'series' with 'show name'

jkoop:
The word 'series' is used in Britain like America's 'season'. So avoiding the use of the word 'series' avoids possible confusion.

### DIFF
--- a/general/server/media/shows.md
+++ b/general/server/media/shows.md
@@ -5,18 +5,18 @@ title: Shows
 
 # Shows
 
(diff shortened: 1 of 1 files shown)


EraYaN:
This should be more than clear from context, given that the word season is used throughout the sentence and in the example structure.

dkanada:
At the suggestion of @EraYaN I'll close this, since we use series throughout the organization to refer to shows.

PR #424, opened 2020-10-19
Fix Docker Run Command for Mountpoints

Artiume:


general/administration/installing.md
```diff
@@ -91,7 +91,7 @@ The basic steps to create and run a Jellyfin container using Docker are as follo
    docker run -d \
     --volume /path/to/config:/config \
     --volume /path/to/cache:/cache \
-    --volume /path/to/media:/media \
+    --mount type=bind,source=/path/to/media,target=/media \
```
nielsvanvelzen, line 94: The documentation below this snippet still uses `--volume`. Additionally it might be useful to specify why the two mounts use the `--volume` argument instead of `--mount`.

Artiume: thanks, i updated it.

nielsvanvelzen (approved):

Artiume:
I did a little rearranging, shouldn't need reapproval.

general/administration/installing.md
PrplHaz4: The way the volumes are specified (with a leading `/`) makes them bind mounts, even though the doc says they are docker named volumes.

IMO, this doc shouldn't reference docker named volumes at all - most of the projects I follow and use do not mention named volumes - it is just an added layer of confusion here, leading to lots of "my config went poof" issues.

Maxr1998 (approved):

PR #427, opened 2020-10-23
Updated the list of supported Debian releases.

ddjxie:


joshuaboniface:
We do support Bullseye. The problem is like any Unstable release, it's a moving target; hence the issue you see with Jellyfin FFMpeg. We'll have to rebuild that with the newer library version. I can do so this weekend.

dkanada (approved):

PR #431, opened 2020-10-29
Distinguish between Samsung Legacy & Tizen client

HimbeersaftLP:
Previously it still linked to the deprecated repo even though the heading said Tizen.

I added a separate heading for the legacy Jellyfin client and updated the link. I also added some information to distinguish between them.

### DIFF
--- a/general/clients/index.md
+++ b/general/clients/index.md
@@ -9,6 +9,8 @@ Clients connect your devices to your Jellyfin server and let you view your conte
 
 > [!NOTE]
 > Many clients have no active maintainers at the moment. If you interested in helping out, please see our [contribution guide](xref:contrib-index) and feel free to contact us for more information!
+>
+> They aren't on this page, some of them can be found at the [jellyfin-archive organization on GitHub](https://github.com/jellyfin-archive).
 
 Do you have a client that interfaces with Jellyfin and want to see it listed here? Please [submit a pull request](https://github.com/jellyfin/jellyfin-docs)!
 
--- a/general/clients/index.md
+++ b/general/clients/index.md
@@ -115,16 +115,6 @@ The official Jellyfin Roku app.
 
 ## Jellyfin Cross-Platform Clients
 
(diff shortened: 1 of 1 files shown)


dkanada:
I don't think we should add a link to the deprecated version at all, since it can't even be installed on older devices and is basically useless at this point.

HimbeersaftLP:
Oh sorry, I didn't know that, I thought it would work because of [that issue comment](https://github.com/jellyfin-archive/jellyfin-samsungtv/issues/6#issuecomment-615904568).

In that case, should the Chrome app also be removed?

dkanada:
Yeah, the Chrome app is unmaintained so we should probably remove it as well.

HimbeersaftLP:
Perhaps include a section at the bottom like "Some old, unmaintained or unfinished clients can be found at https://github.com/jellyfin-archive"?

HimbeersaftLP:
I just noticed that [jellyfin-desktop has also been archived](https://github.com/jellyfin-archive/jellyfin-desktop/commit/f69a32dd993d1a692c4c872684190ec911c78f4b) recently, also @chaosinnovator's [jellyfin-xamarin](https://github.com/chaosinnovator/jellyfin-xamarin) didn't have any commits since December 2019, should they be removed as well?

Also Web Scrobbler isn't really a Jellyfin client, just a general browser extension, so I'm not sure if that should be on that page either.

HimbeersaftLP:
Perhaps there should be a separate documentation page for those 🤔

dkanada:
That could work! It can be done in a separate pull request though.

dkanada (approved):

HimbeersaftLP:
Then again, what info would even be on such a docs page that the repo @jellyfin-archive org doesn't already convey, unless I also search around GitHub for 3rd party unmaintained clients. In which case I run into the problem: At which grade of completion does it deserve to be on said page? I think the notice on top is good enough. What would you say?

PR #432, opened 2020-10-30
Adding docs for multiple kodi clients

phyzical:
Hey there!

Quick rundown of my issue found [here](https://www.reddit.com/r/jellyfin/comments/jk77pk/jellyfin_kodi_question/
)  
i ran into an issue with multiple kodi clients and the way i initial set them up so figured id make a quick pr to hopefully stop future users from doing the same :+1: 
let me know if the wording isnt correct and i can update accordingly

### DIFF
--- a/general/clients/kodi.md
+++ b/general/clients/kodi.md
@@ -113,4 +113,4 @@ The Jellyfin for Kodi addon doesn't natively handle multiple user accounts.  For
 
 ## Multiple Clients
 
-When using multiple kodi clients do not copy kodi's db (i.e myvideosXYZ.db, jellyfin.db) files from one client to the other to try and reduce initial syncing time. This will work, but it will confuse jellyfin and all the clients will be seen as the same client causing syncing between these clients not to work correctly.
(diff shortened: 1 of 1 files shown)


general/clients/kodi.md
```diff
@@ -110,3 +110,7 @@ The Jellyfin for Kodi addon doesn't natively handle multiple user accounts.  For
 
 > [!NOTE]
 > Kodi's default skin does not display all unicode characters. To display unicode characters the skin's font must be changed.
+
+## Multiple Clients
+
+When using multiple kodi clients do not copy kodi's db (i.e myvideosXYZ.db, jellyfin.db) files from one client to the other to try and reduce initial syncing time. This will work, but it will confuse jellyfin and all the clients will be seen as the same client causing syncing between these clients not to work correctly.
```
dkanada, line 116:
```suggestion
When using multiple Kodi clients do not copy Kodi's database (i.e. myvideosXYZ.db, jellyfin.db) files from one client to the other to try and reduce initial syncing time. This will work, but it will confuse Jellyfin and all the clients will be seen as the same client, causing sync between these clients not to work correctly.
```

dkanada (approved):

PR #434, opened 2020-11-04
feature/adding-docs-for-multiple-kodi-clients

phyzical:
found a workaround to the initial issue that cause me to make the first pr 
so refactored multiple clients documentation

### DIFF
--- a/general/clients/kodi.md
+++ b/general/clients/kodi.md
@@ -114,7 +114,7 @@ The Jellyfin for Kodi addon doesn't natively handle multiple user accounts.  For
 
 ## Multiple Clients
 
-When using multiple kodi clients it is not recommended to copy files from one client to the other to try and reduce initial syncing time. This will work, but it will confuse jellyfin and all the clients will be seen as the same client causing syncing between these clients not to work correctly.
+When using multiple Kodi clients it is not recommended to copy files from one client to the other to reduce initial syncing time. This will work, but it will confuse Jellyfin and all the clients will be seen as the same client causing sync between these clients not to work correctly.
 
 If you find you cannot perform a full scan on a certain device though please try the steps below.
 
--- a/general/clients/kodi.md
+++ b/general/clients/kodi.md
@@ -113,4 +113,26 @@ The Jellyfin for Kodi addon doesn't natively handle multiple user accounts.  For
 
 ## Multiple Clients
 
-When using multiple kodi clients do not copy kodi's db (i.e myvideosXYZ.db, jellyfin.db) files from one client to the other to try and reduce initial syncing time. This will work, but it will confuse jellyfin and all the clients will be seen as the same client causing syncing between these clients not to work correctly.
+When using multiple kodi clients it is not recommended to copy files from one client to the other to try and reduce initial syncing time. This will work, but it will confuse jellyfin and all the clients will be seen as the same client causing syncing between these clients not to work correctly.
+
+If you find you cannot perform a full scan on a certain device though please try the steps below.
+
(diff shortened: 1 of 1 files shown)


oddstr13:
Is the initial sync really slow enough to warrant this amount of manual work to reduce it?
In any case, it would be the client GUID confusing the server, as this should always be unique.
Copying everything and then deleting this file on the target *may* work. I have not tested this.
`~/.kodi/userdata/addon_data/plugin.video.jellyfin/jellyfin_guid`

phyzical:
(for me) it was less about how slow it was and more the fact my RPI would crash. (there was a fix for ram leaks which allowed for library scans to progress, but then i started having something else cause the sync to kill the system, just assume its pushing the pi too hard (~1500 movies and ~1500 tv shows)  it takes about 40 mins on a non constrained system)

but your right for the important part is just the guid, just figured if the steps are verbose AF cant go wrong

general/clients/kodi.md
```diff
@@ -114,4 +114,26 @@ The Jellyfin for Kodi addon doesn't natively handle multiple user accounts.  For
...
+4. Reboot client 2.
+5. You should be prompted to enter login details again, but a full scan should not be required.
+6. Verification.
+   * Reboot client 1 and 2. Both of these clients should auto login.
+   * Load up the jellyfin web dashboard -> devices page and client 1 and client 2 should have separate squares.
+
+>* Note: For this to work client 1 and 2 should be running the same kodi major versions.
+>* Note: For this whole process to be working correctly you should see a square per client.
+>* If you do notice that only 1 square is present  try rebooting each client one at a time.
+>* If you do see that they are "fighting for a slot" then jellyfin views them both as the same.
```
mcarlton00, line 138: This is a bit unclear where a user would be looking for the squares since it's separate from the mention of the devices page above.  It also doesn't say how to identify the "fighting for a slot".

general/clients/kodi.md
```diff
@@ -114,4 +114,26 @@ The Jellyfin for Kodi addon doesn't natively handle multiple user accounts.  For
...
+   * Remove KODI_FOLDER/userdata/addon_data/plugin.video.jellyfin/jellyfin_guid from client 2.
+4. Reboot client 2.
+5. You should be prompted to enter login details again, but a full scan should not be required.
+6. Verification.
+   * Reboot client 1 and 2. Both of these clients should auto login.
+   * Load up the jellyfin web dashboard -> devices page and client 1 and client 2 should have separate squares.
+
+>* Note: For this to work client 1 and 2 should be running the same kodi major versions.
+>* Note: For this whole process to be working correctly you should see a square per client.
+>* If you do notice that only 1 square is present  try rebooting each client one at a time.
```
mcarlton00, line 137:
```suggestion
>* If you do notice that only 1 square is present, try rebooting each client one at a time.
```

mcarlton00 (commented):
I can't say that I've tested this, but it sounds like something that's easy for mistakes or weirdness to happen.  It's also not as clear as it could be.

ie: `myvideosxyz.db` vs `MyVideosXXX.db`.  In most OS's, capitalization matters.  Better to be more explicit about these things, imo.

There's also a slim chance for unintended side effects, if the library on one server has been modified a bunch and then gets copied to another, there's a slight chance for row ID weirdness.  I can't say that with any confidence, but I've seen far stranger things come out of Kodi sometimes.

dkanada:
@phyzical did you plan on continuing this pull request?

phyzical:
@dkanada, sorry i decided that i agreed with mcarlton's response that it may cause more harm than good to have as a user guide approuch, i will close it now 👍

PR #470, opened 2021-01-01
Split Anime plugin

Artiume:


general/server/plugins/index.md
```diff
@@ -105,6 +143,14 @@ Scrape poster images for movies, shows, and artists in your library from [fanart
 
 * [GitHub](https://github.com/jellyfin/jellyfin-plugin-fanart)
 
+#### IPTV
+
+[![Language](https://img.shields.io/github/languages/top/jellyfin/jellyfin-plugin-iptv.svg)](https://github.com/jellyfin/jellyfin-plugin-iptv)
+[![Contributors](https://img.shields.io/github/contributors/jellyfin/jellyfin-plugin-iptv.svg)](https://github.com/jellyfin/jellyfin-plugin-iptv)
+[![License](https://img.shields.io/github/license/jellyfin/jellyfin-plugin-iptv.svg)](https://github.com/jellyfin/jellyfin-plugin-iptv)
+
+Plugin to provide additional IPTV support.
```
dkanada, line 152: Hasn't this one been archived?

Artiume: i dunno,  never used it lol. I'll check and fix it

dkanada: I recently removed it because I thought that was the case, if it's still in use then I was wrong.

Artiume: i saw that it was in jelly-archive but yeah  afaik, it was in use. i never heard anything about it being deprecated. it was in 10.6.4's manifest.

dkanada: We talked about the deprecation on Matrix, apparently it's no longer required, but I didn't ask why.

general/server/plugins/index.md
```diff
@@ -45,17 +45,55 @@ All plugins hosted on the repository can be built from source and manually added
 
 ### Official Plugins
 
-#### Anime
+#### Anime Plugins
```
dkanada, line 48: I don't think we need a special category just for anime plugins. I wouldn't mind it, but it needlessly indents all the plugins one header lower.

Artiume: i had set it uo this way because kitsu is the oddball out with its name. i could format the page to match the webui, making categories for each type. or just make it normal.

dkanada: Why is Kitsu the oddball? The plugins have descriptions right below the titles to explain their purpose, so I'm not sure a category is needed.

oddstr13: If anything, it would make sense to group them by category, so Metadata plugins rather than Anime plugins.
With that said, it shouldn't be too much work to whip up a dynamic page that pulls the info from the plugin repo itself, replacing this whole section.

oddstr13 (approved):

PR #476, opened 2021-01-04
how to uninstall macos

Artiume:


anthonylavado (approved):

anthonylavado:
Should probably add a part about dragging the .app to the trash, but this will do.

PR #493, opened 2021-02-02
fixed comment in the apache reverse proxy guide

stefanvi:
I believe line 50 in the apache reverse proxy page should be commented. Otherwise it is not valid syntax and apache will throw the following error:
`
AH00526: Syntax error on line 40 of /etc/apache2/sites-enabled/000-default.conf:
`
`
Invalid command 'Disable', perhaps misspelled or defined by a module not included in the server configuration
`

Commenting this line out fixes the issue for me.


nielsvanvelzen (approved):

nielsvanvelzen:
Thanks!

PR #500, opened 2021-02-19
Updated the podman section in installing.md 

alexpattyn:
Updated the podman section, so the container is a little less exposed by default. I.e. `--privileged` is removed and only the ports needed are passed. 

Then I added a subsection on managing the service with systemd. Along with some sandboxing options that have been tested to work, since that can involve a lot of guess work.


Shadowghost (approved):
Can't test the systemd additions, but seems fine to me.

ThibaultNocchi:
Could you squash your 3 commits together @alexpattyn please? :)

alexpattyn:
@ThibaultNocchi Will do. I have to make some changes to the systemd section first (planning on taking a look at that this weekend).

Artiume:
Do you still need to make any changes?

alexpattyn:
@Artiume Just have to squash the commits. Dumb question but can one do that in github directly?

Artiume:
not sure from your end. i can merge and squash but i think it'll say that it's my commit, if you dont mind i can take care of it for you

alexpattyn:
Yeah go ahead. I'll have to figure that out another time.

PR #507, opened 2021-03-09
Correct web socket nginx proxy config

tonyfinn:
Without the /socket, syncplay will not work on jellyfin-web in 10.7

### DIFF
--- a/general/networking/nginx.md
+++ b/general/networking/nginx.md
@@ -88,7 +88,7 @@ server {
 
     location /socket {
         # Proxy Jellyfin Websockets traffic
-        proxy_pass http://$jellyfin:8096;
+        proxy_pass http://$jellyfin:8096/socket;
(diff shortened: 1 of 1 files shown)


ThibaultNocchi (approved):

nielsvanvelzen:
I'm using the old config in my nginx configuration and synplay with 10.7 works fine.

ThibaultNocchi:
According to the doc, it should do the same thing: https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/

tonyfinn:
Ah, good catch. I was basing this on the issues I had with my setup on upgrading to 10.7, but actually it looks like the real issue was that

```
proxy_pass http://jellyfin
```

and


```
proxy_pass http://jellyfin/
```

are not the same thing.

PR #512, opened 2021-03-18
Fixed config file typo in Nginx

ThibaultNocchi:


general/networking/nginx.md
```diff
@@ -25,7 +25,7 @@ server {
     # return 301 https://$host$request_uri;
 #}
 
-#server {
+server {
```
nielsvanvelzen, line 28: I think the line above this one should be uncommented too

PR #516, opened 2021-03-24
Document the /metrics Prometheus endpoint

strugee:
I made a new page specifically for monitoring and moved the `/health` endpoint documentation into it as well as this new Prometheus documentation. (I'm kinda lumping metrics and monitoring together, but that seems fine.)

### DIFF
--- a/general/networking/monitoring.md
+++ b/general/networking/monitoring.md
@@ -1,0 +1,30 @@
+---
+uid: monitoring
+title: Monitoring
+---
+
+## Monitoring
+
+Jellyfin has two monitoring and metrics endpoints built-in: a basic health check endpoint and a Prometheus-compatible metrics endpoint.
+
+### Health check endpoint
+
+Jellyfin exposes the `/health` endpoint designated for checking the status of the underlying service. Currently this will verify HTTP and database connectivity and return a `200 OK` response if successful. You can see this for yourself by using `curl`:
+
+```sh
+curl -i http://myserver:8096/health
+```
+
+The `-i` option tells `curl` to also print the HTTP response code and headers.
+
+### Prometheus metrics
+
+Jellyfin can make [Prometheus](https://prometheus.io/) metrics available at `/metrics`, but this is turned off by default to avoid unintentionally leaking this information on the public internet. To enable it, you will need to edit `/etc/jellyfin/system.xml` and change this line:
+
+```xml
+<EnableMetrics>false</EnableMetrics>
+```
+
+to say `true` instead of `false`.
+
+If you have a [reverse proxy](xref:network-index#running-jellyfin-behind-a-reverse-proxy) configured, you can configure it to block access to the `/metrics` endpoint except for your internal network.
--- a/general/networking/monitoring.md
+++ b/general/networking/monitoring.md
@@ -19,12 +19,10 @@ The `-i` option tells `curl` to also print the HTTP response code and headers.
 
 ### Prometheus metrics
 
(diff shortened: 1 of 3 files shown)


general/networking/monitoring.md
```diff
@@ -0,0 +1,30 @@
...
+curl -i http://myserver:8096/health
+```
+
+The `-i` option tells `curl` to also print the HTTP response code and headers.
+
+### Prometheus metrics
+
+Jellyfin can make [Prometheus](https://prometheus.io/) metrics available at `/metrics`, but this is turned off by default to avoid unintentionally leaking this information on the public internet. To enable it, you will need to edit `/etc/jellyfin/system.xml` and change this line:
+
+```
```
Shadowghost:
```suggestion
```xml
```

general/networking/monitoring.md
```diff
@@ -0,0 +1,30 @@
...
+
+Jellyfin can make [Prometheus](https://prometheus.io/) metrics available at `/metrics`, but this is turned off by default to avoid unintentionally leaking this information on the public internet. To enable it, you will need to edit `/etc/jellyfin/system.xml` and change this line:
+
+```xml
+<EnableMetrics>false</EnableMetrics>
+```
+
+to say `true` instead of `false`.
+
+If you have a [reverse proxy](xref:network-index#running-jellyfin-behind-a-reverse-proxy) configured, you can configure it to block access to the `/metrics` endpoint except for your internal network.
```
Shadowghost, line 30: We should document how to secure this endpoint in the reverse proxy configs we provide (or just block it per default in them). Not needed for this PR to get merged though, just something we need to consider documenting properly.
I'm open for opinions on how to deal with this.

ThibaultNocchi (approved):
LGTM

general/networking/monitoring.md
```diff
@@ -0,0 +1,30 @@
...
+
+```sh
+curl -i http://myserver:8096/health
+```
+
+The `-i` option tells `curl` to also print the HTTP response code and headers.
+
+### Prometheus metrics
+
+Jellyfin can make [Prometheus](https://prometheus.io/) metrics available at `/metrics`, but this is turned off by default to avoid unintentionally leaking this information on the public internet. To enable it, you will need to edit `/etc/jellyfin/system.xml` and change this line:
```
Shadowghost, line 22:
```suggestion
Jellyfin can make [Prometheus](https://prometheus.io/) metrics available at `/metrics`, but this is turned off by default to avoid unintentionally leaking this information on the public internet. To enable it, you will need to edit `/etc/jellyfin/system.xml` and change this line from `false` to `true`:
```

general/networking/monitoring.md
```diff
@@ -0,0 +1,30 @@
...
+
+```xml
+<EnableMetrics>false</EnableMetrics>
+```
+
+to say `true` instead of `false`.
```
Shadowghost, line 28:
```suggestion
```

Shadowghost (approved):

general/networking/monitoring.md
strugee: markdownlint caught this - I must've pushed while you already had the tab open.

general/networking/monitoring.md
strugee: Maybe each reverse proxy config page could just have a section at the end like, "to block the metrics endpoint add this snippet"?

I'm not convinced that this is a huge problem though. If you're advanced enough to be running Prometheus and actively enabling Jellyfin's metrics support, you probably also know enough to just write this configuration on your own, based on your reverse proxy's docs. If anyone else buys that argument I can add something to this PR saying "consult your reverse proxy's documentation for how to do this" - then we could avoid cluttering all the individual reverse proxy pages, which are going to be needed/consulted the most by beginner administrators anyway. I don't feel strongly about it though.

strugee:
CI is passing.

Maxr1998 (approved):

strugee:
Thank you!

PR #519, opened 2021-03-29
Added VSC lint on save

ThibaultNocchi:


Shadowghost (approved):

Maxr1998 (approved):

.vscode/settings.json
```diff
@@ -1,5 +1,9 @@
...
-        "src/**": true
-    }
-}
\ No newline at end of file
+{
+    "search.exclude": {
+        "src/**": true
+    },
+    "editor.codeActionsOnSave": {
+        "source.fixAll.markdownlint": true
```
Maxr1998, line 6: What happens if the markdownlint plugin is not installed?

ThibaultNocchi: Just tested, and nothing happens. Like no error or anything. Also the extension is suggested by our environment so I'd go for it

Maxr1998 (approved):

PR #522, opened 2021-04-06
Add warning and example about api_key parameters in logfiles

GermanCoding:
Relates to this issue: https://github.com/jellyfin/jellyfin/issues/5415

This tells users that they should be aware about what they're logging and for nginx only it includes an example on how to filter such information. I do not know the other webservers well enough to write examples for these from scratch.

### DIFF
--- a/general/networking/index.md
+++ b/general/networking/index.md
@@ -79,7 +79,7 @@ It's possible to run Jellyfin behind another server acting as a reverse proxy.  
 > These examples assume you want to run Jellyfin under a sub-domain (e.g. jellyfin.example.com), but are easily adapted for the root domain if desired.
 
 > [!WARNING]
-> Be careful when logging requests with your reverse proxy. Jellyfin sometimes sends authentication information as part of the URL, so logging the full request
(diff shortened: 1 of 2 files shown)


general/networking/index.md
```diff
@@ -78,6 +78,11 @@ It's possible to run Jellyfin behind another server acting as a reverse proxy.
 > In order for a reverse proxy to have the maximum benefit, you should have a publically routable IP address and a domain with DNS set up correctly.
 > These examples assume you want to run Jellyfin under a sub-domain (e.g. jellyfin.example.com), but are easily adapted for the root domain if desired.
 
+> [!WARNING]
+> Be careful when logging requests with your reverse proxy. Jellyfin sometimes sends authentication information as part of the URL, so logging the full request
```
PrplHaz4, line 82: maybe include a list of known secrets that should be filtered in this line?
> Jellyfin sometimes sends authentication information as part of the URL (e.g. `api_key`), so logging the full request can expose these secrets in your logfile.

general/networking/nginx.md
```diff
@@ -170,6 +170,27 @@ server {
 
 ## Extra Nginx Configurations
 
+## Censor sensitive information in logs
```
Artiume, line 173:
```suggestion
### Censor sensitive information in logs
```

Artiume (approved):

PR #523, opened 2021-04-09
Add Finamp

UnicornsOnLSD:
Finamp is my new music player for Jellyfin. For now, it is only available on Android, but I should be able to release it on iOS soon.

It should probably also be added to https://jellyfin.org/clients/.


mcarlton00 (approved):
Looks nice, thanks.  Welcome to the party

PR #526, opened 2021-04-14
Add Jellyfin Media Player.

iwalton3:
This adds Jellyfin Media Player to the client list.


ThibaultNocchi (commented):
I'd add that it supports a lot of codecs or something like this, not all people know what MPV is and the support it brings :)

iwalton3:
I added a note that it direct plays most formats.

ThibaultNocchi (approved):

Maxr1998 (approved):

PR #531, opened 2021-05-01
feat: Point users to Extrepo based install for Debian

ypid:
The [suggested way to install Jellyfin on Debian](https://jellyfin.org/docs/general/administration/installing.html#debian) involves downloading the signing key from the same webserver that the repo is served from without pinning the key by fingerprint or the like.

[Extrepo](https://salsa.debian.org/extrepo-team/extrepo) exists to address this problem. It has the advantage of being in Debian (and Ubuntu) so there is a trust chain for the user that can be publicly reviewed.

This PR proposes switching to an Extrepo based install method. Note that [I added Jellyfin to extrepo-data](https://salsa.debian.org/extrepo-team/extrepo-data/-/merge_requests/72) so you might need to discuss if Extrepo is suitable for you here as Extrepo is unknown so far in the Jellyfin community as far as I can tell (and I have not mentioned it earlier).

* https://grep.be/blog/en/computer/debian/Announcing_extrepo/

### DIFF
--- a/general/administration/installing.md
+++ b/general/administration/installing.md
@@ -591,10 +591,10 @@ The Jellyfin team provides a Debian repository for installation on Debian Stretc
 
 Steps 1 to 3 can also be replaced by:
 
-    ```sh
-    sudo apt install extrepo
-    sudo extrepo enable jellyfin
-    ```
+```sh
+sudo apt install extrepo
+sudo extrepo enable jellyfin
+```
 
 1. Install HTTPS transport for APT as well as `gnupg` and `lsb-release` if you haven't already.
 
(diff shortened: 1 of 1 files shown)


Artiume:
While it would make the process easier for some. I will say I'll occasionally reference this page specifically for the key. Could we make this an alternative means instead of entirely removing the original method

ypid:
Sounds good to me.

Artiume (approved):

Artiume:
Looks like the linter failed because the indentation of L594-597 should match L592.

https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md046---code-block-style

PR #532, opened 2021-05-02
add codec profile tables

Artiume:


ThibaultNocchi:
Maybe I'd add a single sentence to say it also detects your browser compatibility?

Artiume:
Updated. I have a feeling someone will hit us up one day saying that table supports X but jf is transcoding. cie la vie

ThibaultNocchi (approved):
Oh yeah true indeed, but that won't change a lot from basic "it doesn't work" and will help a little more people I guess ^^

PR #559, opened 2021-08-15
Update shows.md

briandipalma:
I'll raise one for movies if this is OKed. Can a developer verify the syntax, I'm not a C# dev so I may be wrong about the syntax, the code is in MovieResolver.cs.

### DIFF
--- a/general/server/media/shows.md
+++ b/general/server/media/shows.md
@@ -5,11 +5,11 @@ title: Shows
 
 # Shows
 
-The most common naming scheme for shows is categorizing the files by series and then season. Another common method is simply using series folders, especially for shows that are organized by air date and those without seasons. Adding the year at the end in parentheses will yield the best results when scraping metadata.
+The most common naming scheme for shows is categorizing the files by series and then season. Another common method is simply using series folders, especially for shows that are organized by air date and those without seasons. Adding the year at the end in parentheses along with a media DB ID will yield the best results when scraping metadata. The two DBs supported by Jellyfin are [IMDB](https://www.imdb.com/) and [TMBD](https://www.themoviedb.org/). Media management software such as [Sonarr](https://sonarr.tv/) can be configured to add these IDs in to the folder names. The examples below show the ID syntax `[idtype-ID]`.
 
 ```txt
 Shows
-├── Series (2010)
+├── Series (2010) [imdbid-ID]
 │   ├── Season 01
 │   │   ├── Episode S01E01-E02.mkv
 │   │   ├── Episode S01E03.mkv
@@ -17,7 +17,7 @@ Shows
 │   └── Season 02
 │       ├── Episode S02E01.mkv
 │       └── Episode S02E02.mkv
-└── Series (2018)
+└── Series (2018) [tmdbid-ID]
     ├── Episode S01E01.mkv
     ├── Episode S01E02.mkv
     ├── Episode S02E01-E02.mkv
--- a/general/server/media/shows.md
+++ b/general/server/media/shows.md
@@ -5,11 +5,13 @@ title: Shows
 
 # Shows
 
-The most common naming scheme for shows is categorizing the files by series and then season. Another common method is simply using series folders, especially for shows that are organized by air date and those without seasons. Adding the year at the end in parentheses along with a media DB ID will yield the best results when scraping metadata. The two DBs supported by Jellyfin are [IMDB](https://www.imdb.com/) and [TMBD](https://www.themoviedb.org/). Media management software such as [Sonarr](https://sonarr.tv/) can be configured to add these IDs in to the folder names. The examples below show the ID syntax `[idtype-ID]`.
+The most common naming scheme for shows is categorizing the files by series and then season. Another common method is simply using series folders, especially for shows that are organized by air date and those without seasons. Adding the year at the end in parentheses will yield the best results when scraping metadata.
+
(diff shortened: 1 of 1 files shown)


TRaSH-:
it would be nice if Jellyfin would sort of use the same identifying what plex uses,
this makes it easier for people that want to switch or just run both.

briandipalma:
For what it's worth it seem like Plex/Emby and Jellyfin all use slightly different syntax/keys. Not ideal but I guess it should be straightforward to support all of them.

briandipalma:
I'm not a C# dev but I can try and take a look at this if it's OK for Jellyfin to support them all?

general/server/media/shows.md
```diff
@@ -5,19 +5,19 @@ title: Shows
 
 # Shows
 
-The most common naming scheme for shows is categorizing the files by series and then season. Another common method is simply using series folders, especially for shows that are organized by air date and those without seasons. Adding the year at the end in parentheses will yield the best results when scraping metadata.
+The most common naming scheme for shows is categorizing the files by series and then season. Another common method is simply using series folders, especially for shows that are organized by air date and those without seasons. Adding the year at the end in parentheses along with a media DB ID will yield the best results when scraping metadata. The two DBs supported by Jellyfin are [IMDB](https://www.imdb.com/) and [TMBD](https://www.themoviedb.org/). Media management software such as [Sonarr](https://sonarr.tv/) can be configured to add these IDs in to the folder names. The examples below show the ID syntax `[idtype-ID]`.
 
 ```txt
 Shows
-├── Series (2010)
+├── Series (2010) [imdbid-ID]
```
MrTimscampi, line 12: This makes it seem like you *have* to put the IDs in the file names for proper naming, which isn't the case.

Sidenote: If anything, it may lead to more false positives for people, as it's more stuff for the filename parser to deal with and try to recognize. Hence why I'd argue that this feature should be removed altogether and replaced with support for [Kodi's Parsing NFOs](https://kodi.wiki/view/NFO_files/Parsing).

general/server/media/shows.md
```diff
@@ -5,19 +5,19 @@ title: Shows
 
 # Shows
 
-The most common naming scheme for shows is categorizing the files by series and then season. Another common method is simply using series folders, especially for shows that are organized by air date and those without seasons. Adding the year at the end in parentheses will yield the best results when scraping metadata.
+The most common naming scheme for shows is categorizing the files by series and then season. Another common method is simply using series folders, especially for shows that are organized by air date and those without seasons. Adding the year at the end in parentheses along with a media DB ID will yield the best results when scraping metadata. The two DBs supported by Jellyfin are [IMDB](https://www.imdb.com/) and [TMBD](https://www.themoviedb.org/). Media management software such as [Sonarr](https://sonarr.tv/) can be configured to add these IDs in to the folder names. The examples below show the ID syntax `[idtype-ID]`.
```
MrTimscampi, line 8: Avoid "jargon" and needless contractions/abbreviations like "ID" and "DB". It makes it more difficult for the average user to read and understand the documentation properly.

This also feels more like something that should live in a `tip` block, instead of in the main description.

Something like:

```markdown
> [!TIP]
> In order to help with identifying series, Jellyfin can make use of a media provider identifier.
> This can be specified in your show's folder name, for example as: `Series (2010) [tt0106145]`
```

Also, we avoid any references to software used for piracy, such as Sonarr. If you really want to add a note to replace that reference, something like "Consult your media manager of choice's documentation for ways to include this identifier automatically", but I don't even think that's really necessary, as it's outside the scope of use of Jellyfin itself.

MrTimscampi (changes requested):

general/server/media/shows.md
briandipalma: That's much better, replaced what I wrote with your suggestion.

general/server/media/shows.md
briandipalma: Removed that change. Although I'd argue that parsing `[imdbid-%ID%]` should be relatively doable and bug free. I also thought Jellyfin already parsed NFOs? Also NFOs can contain more data that can become stale and some people would prefer for Jellyfin not to have write access to the media files. Although maybe NFOs make more sense for larger collections as I imagine scanning them becomes quicker?

briandipalma:
It appears that `[imdbid=tt8001226]` is also supported based on Reddit posts. Should I update the PR to reflect that?

Artiume:
Only add it if you can verify the functionality. That said, whenever wrong info is put in the documents, it isn't the end of the world, people are pretty quick to point it out if it's egregious enough lol.

general/server/media/shows.md
```diff
@@ -7,6 +7,8 @@ title: Shows
 
 The most common naming scheme for shows is categorizing the files by series and then season. Another common method is simply using series folders, especially for shows that are organized by air date and those without seasons. Adding the year at the end in parentheses will yield the best results when scraping metadata.
 
+[!TIP] In order to help with identifying series, Jellyfin can make use of a media provider identifier. This can be specified in your show's folder name, for example as: `Series (2010) [imdbid-tt0106145]` or `Series (2018) [tmdbid-65567]`
```
Artiume, line 10:
```suggestion
[!TIP] In order to help with identifying a series, Jellyfin can make use of media provider identifiers. This can be specified in your show's folder name, for example: `Series (2010) [imdbid-tt0106145]` or `Series (2018) [tmdbid-65567]`
```

briandipalma:
> Only add it if you can verify the functionality. That said, whenever wrong info is put in the documents, it isn't the end of the world, people are pretty quick to point it out if it's egregious enough lol.

Yeah it worked, I used it to fix a problem identifying a TV series.

Artiume (approved):

PR #562, opened 2021-08-27
Add Finamp Play Store, App Store, F-Droid links

UnicornsOnLSD:
Forgot to update this page when I added Finamp to the stores, so I've added the links.


ThibaultNocchi:
I think I'd put the client in the cross-platform section now that it's not only on Android :)

Artiume (approved):

UnicornsOnLSD:
Moved to cross-platform, removing the Xamarin client should probably be done in a different PR/commit

PR #568, opened 2021-09-03
Add note about the different kodi add-ons

Artiume:


Shadowghost (changes requested):

Shadowghost (approved):

mcarlton00:
Little late now, but I'd just like the point out that both of those descriptions are already on this page in the respective section about each addon further down the page.  You can do with this info what you'd like.

PR #574, opened 2021-09-07
Update fail2ban.md

solankimihir:
When running Jellyfin and/or Reverse Proxy in a docker-container, then the following variable needs to be defined in jails.local or jails.d/jellyfin.local
chain=DOCKER-USER


solankimihir:
Will rectify the issues with markdownlint and resubmit the pull request

PR #576, opened 2021-09-08
hostmode HDHR

Artiume:


thornbill (approved):

PrplHaz4:
Is this purposely changing back to using docker volumes rather than bind mounts?  I thought they were changed to bind mounts a while ago to keep users from wondering where their data went...

Artiume:
No no, nothing changed. Someone made the comment that it'd make more sense to specify how to do bind mounts first and not volumes so i flipped them

PR #578, opened 2021-10-02
Update build for dotnet 6.0 and npm

1337joe:
Changed source of dotnet - 6.0 not yet available in repo
Removed yarn, added nodejs
Updated build commands to match dockerfiles


1337joe:
Between my learning to use Rider instead of building in the terminal and dotnet 6 not being prerelease anymore this went out of date quickly and I never found the motivation to go back and redo it.

#614 looks like it covers the important bits

PR #584, opened 2021-10-09
How to unlock Blocked User

Xyz00777:
When the admin account is locked out you have only the way over the jellyfin.db with sqlite query to unlock the admin user. At the moment there i no way documented on the internet. The corresponding issue is these: 
https://github.com/jellyfin/jellyfin/issues/6645#event-5412515320
after a few times back and forth cvium and i switched to the matrix chat and got the problem fixed there

### DIFF
--- a/general/administration/troubleshooting.md
+++ b/general/administration/troubleshooting.md
@@ -72,3 +72,34 @@ rm -Rfv ~/.config/jellyfin
 rm -Rfv ~/.cache/jellyfin
 rm -Rfv ~/.local/share/jellyfin
 ```
+
+## Unlock locked user account
+
+When the admin account is locked out and the Forgot Password feature is not working, you have to unlock the user manually. To do that, you need to find the `jellyfin.db` file on your system. The default location on Linux is: `/var/lib/jellyfin/data/`.
+For paths in other environments, see [server paths](https://jellyfin.org/docs/general/administration/configuration.html#server-paths).
+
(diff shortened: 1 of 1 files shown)


Xyz00777:
i could bet that i did something wrong for sure or didn't pay attention, if so i'm sorry for that, it's my first push on github :)

KucharczykL:
I think making the directions more generic would be useful. For example, the database file is located in the data folder but that folder can be in many different places depending on how is Jellyfin installed. Linking to https://jellyfin.org/docs/general/administration/configuration.html#server-paths might be better.

KucharczykL:
> > I think making the directions more generic would be useful. For example, the database file is located in the data folder but that folder can be in many different places depending on how is Jellyfin installed. Linking to https://jellyfin.org/docs/general/administration/configuration.html#server-paths might be better.
> 
> @KucharczykL yes we can do that but then i would write these path i found to the linked site from you, because the environment variables dont apply correct every time. eg when i try to use these variables on my linux server to find out where jellyfin data directory is, i get no information where it should be, and these path (looks like it) is the default installation directory.

As someone who only used Jellyfin in Docker, I don't know why that would happen. Maybe something too look into. What do you think about leaving the Linux-specific example as it is and just adding something like:

```
For paths in other environments, see https://jellyfin.org/docs/general/administration/configuration.html#server-paths.
```

PR #594, opened 2021-12-02
Update Podman example

jwillikers:
I recently was helping someone get their Jellyfin server up and running using Podman and figured I should update the existing documentation here.
The Podman example is pretty out-of-date.
This PR adds missing steps when running rootless Podman, updates the Podman run command, and uses better, more secure defaults such as running as non-root within the Jellyfin container.
Changes to the Podman run command should properly allow for automatic updates now, as well.

### DIFF
--- a/general/administration/installing.md
+++ b/general/administration/installing.md
@@ -140,7 +140,7 @@ A community project to deploy Jellyfin on Kubernetes-based platforms exists [at 
 
 ### Podman
 
-[Podman](https://podman.io) allows you to run containers as non-root. It's also the offically supported container solution on RHEL and CentOS.
+[Podman](https://podman.io) allows you to run containers as non-root. It's also the officially supported container solution on RHEL and CentOS.
 
 Steps to run Jellyfin using Podman are almost identical to Docker steps:
 
--- a/general/administration/installing.md
+++ b/general/administration/installing.md
@@ -154,12 +154,12 @@ Steps to run Jellyfin using Podman are almost identical to Docker steps:
 
    ```sh
    podman run \
-    --volume jellyfin-config:/config:Z \
-    --volume jellyfin-cache:/cache:Z \
-    --volume jellyfin-media:/media:ro,z \
-    --publish 8096:8096/tcp \
     --label "io.containers.autoupdate=registry" \
     --name myjellyfin \
+    --publish 8096:8096/tcp \
+    --volume jellyfin-cache:/cache:Z \
+    --volume jellyfin-config:/config:Z \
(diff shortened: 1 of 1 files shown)


general/administration/installing.md
```diff
@@ -140,65 +140,52 @@ A community project to deploy Jellyfin on Kubernetes-based platforms exists [at
...
+2. Open the necessary ports in your machine's firewall to permit access to the Jellyfin server from outside the host. This is necessary for rootless Podman because it won't be able to open these automatically. If your distribution uses `firewalld`, the following commands save and load a new firewall rule opening the HTTP port `8096` for TCP connections.  
 
-   ```sh
-   mkdir /path/to/config
-   mkdir /path/to/cache
-   ```
+    ```sh
+    sudo firewall-cmd --add-port=8096/tcp --permanent
+    sudo firewall-cmd --reload
+    ```
```
Shadowghost, line 156: IMHO opening the firewall should be one of the last steps after initially setting up the container.
Accessint it via localhost should be possible without opening the firewall, right?

general/administration/installing.md
```diff
@@ -252,22 +239,28 @@ As always it is recommended to run the container rootless. Therefore we want to
...
 
     At this point the container will only start when the user logs in and shutdown when they log off. To have the container start as the user at first login we'll have to include one more option.
 
     ```sh
-    loginctl enable-linger <username>
+    loginctl enable-linger $USER
     ```
 
-5. Start the service.
+6. To enable Podman auto-updates, enable the necessary Systemd timer.
```
Shadowghost, line 260:
```suggestion
6. To enable Podman auto-updates, enable the necessary systemd timer.
```

general/administration/installing.md
jwillikers: You can definitely access the Jellyfin container on localhost without this. I wanted to add this step since it can be confusing if you're used to Docker / rootfull Podman where this is done for you automatically. I'll go ahead and move it to a later step.

general/administration/installing.md
jwillikers: Nice catch!

general/administration/installing.md
jwillikers: I've placed the firewall step last and tried to word it such that it emphasizes opening firewall ports is only necessary if you wish to access the Jellyfin container from outside the host.

general/administration/installing.md
jwillikers: Fixed!

Shadowghost (approved):

joshuaboniface (approved):

PR #611, opened 2022-01-07
Give recommended way to get XMLTV data

AndrewBreyen:
There is conflicting repos, scripts, etc and lots of outdated documentation on how to do this. By giving an example, users will be guided to a repo with a docker container that is easy to set up, run, and generates reputable guide data.


AndrewBreyen:
Is there a way to get this reviewed?

joshuaboniface (approved):
Looks good to me, that repo seems reliable and I have no issue linking to it.

PR #616, opened 2022-01-13
Add tip to get opencl working on docker

Vitorvlv:
The only way it started working and getting recognized on clinfo was passing the privileged flag to the container. AMD VAAPI + Tonemapping works now. I'm not completely sure on what are the implications of doing this though, so if anyone has more information then please add it.


Shadowghost:
Can you please squash the commits? Rest looks good to go.

Vitorvlv:
Sorry, still figuring git out, apparently I created 2 commits instead of only the 1 squashing them. That's not right, is it?

PR #629, opened 2022-01-23
Update subpath configuration for caddy

DotIN13:
Document modifications of BaseUrl when there is no access to the web interface.

### DIFF
--- a/general/networking/caddy.md
+++ b/general/networking/caddy.md
@@ -61,7 +61,7 @@ You can serve Jellyfin only at a particular base path and not proxy all other re
 
 To do this, first configure Jellyfin to use a base path.
 If you already have access to the web interface, go to `Admin > Networking` and enter a path like `/jellyfin` in the Base URL field.
-If not, you may instead go to `<Configuration_directory>/network.xml` and modify key `<BaseUrl>` to your needs. For references on the directory location, please consult [Configuration Documentation](https://jellyfin.org/docs/general/administration/configuration.html#configuration-directory).
+If not, you may instead go to `<Configuration Directory>/network.xml` and modify the value of `<BaseUrl>` according to your needs. For information on the directory location, please consult the [configuration documentation](https://jellyfin.org/docs/general/administration/configuration.html#configuration-directory).
 You might have to restart the Jellyfin server for this to take effect.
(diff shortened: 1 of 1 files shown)


general/networking/caddy.md
```diff
@@ -60,7 +60,8 @@ reverse_proxy 127.0.0.1:8096
 You can serve Jellyfin only at a particular base path and not proxy all other requests.
 
 To do this, first configure Jellyfin to use a base path.
-Go to `Admin > Networking` and in the Base URL field, enter a path like `/jellyfin`.
+If you already have access to the web interface, go to `Admin > Networking` and in the Base URL field, enter a path like `/jellyfin`.
```
Maxr1998, line 63:
```suggestion
If you already have access to the web interface, go to `Admin > Networking` and enter a path like `/jellyfin` in the Base URL field.
```

general/networking/caddy.md
```diff
@@ -60,7 +60,8 @@ reverse_proxy 127.0.0.1:8096
...
+If not, you may instead go to `<Configuration_directory>/network.xml` and modify key `<BaseUrl>` to your needs. For references on the directory location, please consult [Configuration Documentation](https://jellyfin.org/docs/general/administration/configuration.html#configuration-directory).
```
Maxr1998, line 64:
```suggestion
If not, you may instead go to `<configuration_directory>/network.xml` and modify the value of `<BaseUrl>` according to your needs. For information on the directory location, please consult the [configuration documentation](https://jellyfin.org/docs/general/administration/configuration.html#configuration-directory).
```

Maxr1998 (commented):
Thanks for the helpful tip, however, I'd reword the sentences a bit.

general/networking/caddy.md
DotIN13: Thanks for the kind suggestions, I agree with all of the proposed adjustments except one.

As I read through the documentation again, I realized the way of expressing`<configuration_directory>` that is the most consistent with the current [configuration documentation](https://jellyfin.org/docs/general/administration/configuration.html) is `<Configuration Directory>`.

Maxr1998: Good point, agreed.

Maxr1998 (approved):

PR #633, opened 2022-02-06
Update QSV to include Dockerfile example

Goblinlordx:
**Changes:**
 - Add docker file example for QSV setup to create container with non-free intel driver

### DIFF
--- a/general/administration/hardware-acceleration.md
+++ b/general/administration/hardware-acceleration.md
@@ -351,7 +351,14 @@ To check information about VA-API on your system install and run `vainfo` from t
    sudo apt remove intel-media-va-driver intel-media-va-driver-non-free -y
    ```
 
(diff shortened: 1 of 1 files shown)


Shadowghost:
I'm a bit hesistant to merge this since it will be obsolet if https://github.com/jellyfin/jellyfin-ffmpeg/pull/89 is merged

Goblinlordx:
Hmmm~ I definitely look forward to that being merged. I am not actually sure why it should prevent a merge of this though as it is strictly documentation and it is an example for what is already mentioned. The way I see it, when the description is removed then the example would also be removed at the same time.

Goblinlordx:
superseded by: https://github.com/jellyfin/jellyfin-ffmpeg/pull/89

PR #634, opened 2022-02-11
More detailed documentation on recording post-processing

AndrewBreyen:
Created new page due to the amount of detail/explanation.

Details given on how post-processing works, and one working example utilizing a "runner" shell script which calls the post-process python script.


AndrewBreyen:
Is there a way to get this reviewed?

Shadowghost:
I for myself don't have enough knowledge on the topic to review this for correctness. Apart from that I don't mind greenlighting and merging this PR.

anthonylavado:
I've assigned it to myself since I have a TV tuner and I am in a position to test it. I cleared through some documentation I was working on for the server-macOS repo, so expect movement on this in the next day or so.

anthonylavado (approved):
This is good revision to the document, and will serve as a well used guide for others. Thank you!

AndrewBreyen:
@anthonylavado thanks for merging this! I think this will be super helpful for others. I have noticed, the page is inaccessible via the sidebar, only accessible via direct link (https://jellyfin.org/docs/general/server/live-tv/post-process) 

I'm not super sure how to fix this, I assumed that adding the new UID (uid: server-live-tv-post-process) and it being its own file would do this? Not sure if I missed something here. Let me know if I can help resolve this.


image

AndrewBreyen:
ahh, I didn't know about the TOC entry that has to be added, I added it here: https://github.com/jellyfin/jellyfin-docs/pull/645

PR #639, opened 2022-03-05
Rewrite CI to use GitHub actions

Shadowghost:


.github/workflows/build.yaml
```diff
@@ -0,0 +1,16 @@
...
+  cancel-in-progress: true
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    uses: Shadowghost/jellyfin-docs/.github/workflows/_meta.yaml@ci-rework
```
1337joe, line 14: Shouldn't this (and publish.yaml) point to `jellyfin/jellyfin-docs` instead of `Shadowghost/jellyfin-docs`?

Shadowghost: Yes but to test it I need to reference an existing workflow :)

.github/renovate.json
```diff
@@ -0,0 +1,22 @@
...
+            "automerge": true,
+            "automergeType": "branch"
+        }
+    ],
+    "git-submodules": {
+        "enabled": true
+    },
+    "timezone": "Etc/UTC",
+    "schedule": ["before 2am"]
+}
```
nielsvanvelzen, line 22: Missing a new line

nielsvanvelzen (approved):

.github/renovate.json
Shadowghost: VS Code autmatically removes newlines at the end of json files.

Is it really required?

nielsvanvelzen: We generally have those for all files enabled via the editorconfigs. I think this repository does too.

Shadowghost: You are correct. The weird thing is I have the editorconfig plugin installed on vscode but it seem to either be ignored or overridden by another setting.

Nonetheless I fixed it manually.

crobibero:
Should the `.ci` folder be removed so re disable azure pipelines?

Shadowghost:
Oh right, I forgot that. I'll add a commit.

crobibero (approved):

PR #657, opened 2022-04-17
nginx reverse proxy: allow bigger file upload ⬆️

EV21:
default values is `client_max_body_size 1M`

Without setting this the user who uploads an image will see a loading circle without any error message.
Even the jellyfin logs are not useful. 

nginx error (logfile) is: client intended to send too large body

### DIFF
--- a/general/networking/nginx.md
+++ b/general/networking/nginx.md
@@ -30,6 +30,9 @@ server {
     # listen [::]:443 ssl http2;
     server_name DOMAIN_NAME;
 
+    ## The default `client_max_body_size` is 1M, so this might be not enaugh for some posters, etc.
+    client_max_body_size 20M;
+
     # use a variable to store the upstream proxy
     # in this example we are using a hostname which is resolved via DNS
     # (if you aren't using DNS remove the resolver line and change the variable to point to an IP address e.g `set $jellyfin 127.0.0.1`)
--- a/general/networking/nginx.md
+++ b/general/networking/nginx.md
@@ -30,7 +30,7 @@ server {
     # listen [::]:443 ssl http2;
     server_name DOMAIN_NAME;
 
-    ## The default `client_max_body_size` is 1M, so this might be not enaugh for some posters, etc.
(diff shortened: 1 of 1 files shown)


general/networking/nginx.md
```diff
@@ -30,6 +30,11 @@ server {
     # listen [::]:443 ssl http2;
     server_name DOMAIN_NAME;
 
+    ## The following setting allows you to upload files like images.
+    sendfile on;
```
Shadowghost, line 34: I'd prefer to either remove this or explicitly set it to off since it's not a requirement to make uploads work and breaks uploading on filesystems like ZFS if enabled.

EV21: Thank you for that input. I misunderstood how sendfile actually works.  📖👀
Since it is disabled (off) by default I would prefer to remove these two lines to keep it short and simple. 
If this is a common issue we might add an extra `## Troubleshooting` section for such things.

I do a quick rebase/force-push

Shadowghost (approved):
Looks good to me, but 50MB might be a bit much since we only upload subs or pictures we should maybe change it to something more conservative like 10MB?

EV21:
I recently uploaded a high resolution poster from theposterdb.com at around 8MB so 10MB might be ok but to be safe 20MB would be better. - I initially chose 50 MB to be on the safe side, because I don't know all the functions and limitations of jellyfin and its plugins.

Shadowghost:
Let's go with 20MB to be sure.

general/networking/nginx.md
```diff
@@ -30,6 +30,9 @@ server {
     # listen [::]:443 ssl http2;
     server_name DOMAIN_NAME;
 
+    ## The default `client_max_body_size` is 1M, so this might be not enaugh for some posters, etc.
```
Shadowghost, line 33:
```suggestion
    ## The default `client_max_body_size` is 1M, this might be not enough for some posters, etc.
```

PR #658, opened 2022-04-26
Fix typos on installation documentation

Gnafu:
These are small changes that may be merged into https://github.com/jellyfin/jellyfin-docs/pull/649 by @Shadowghost


Shadowghost:
Included in #649

PR #664, opened 2022-05-12
NVENC docker instruction updates

lifehackerhansol:
The description mentions that the `deploy` section has to be inside the Jellyfin service, however the visual description shows it as a separate service.

### DIFF
--- a/general/administration/hardware-acceleration.md
+++ b/general/administration/hardware-acceleration.md
@@ -247,12 +247,12 @@ Or with docker-compose >1.28, add the `deploy` section to your Jellyfin service:
 services:
   jellyfin:
     image: jellyfin/jellyfin
-  # ... your Jellyfin config
-  deploy:
-    resources:
-      reservations:
-        devices:
-        - capabilities: [gpu]
+    # ... your Jellyfin config
+    deploy:
+      resources:
+        reservations:
+          devices:
+            - capabilities: [gpu]
 ```
 
 There are some special steps when running with the following option:


general/administration/hardware-acceleration.md
```diff
@@ -220,10 +220,12 @@ Prerequisites:
 
 Follow the instructions in the link above to install the NVIDIA Container Toolkit for your Linux distribution.
 
-Start your container by adding this parameter:
+Start your container by adding these parameters:
 
 ```sh
 --gpus all \
+-e NVIDIA_DRIVER_CAPABILITIES=all \
```
crobibero, line 227: This is already included
https://github.com/jellyfin/jellyfin-metapackages/blob/e78e91625c1292c86d8a604827f6acccb5b77f61/Dockerfile.amd64#L23

general/administration/hardware-acceleration.md
```diff
@@ -232,6 +234,8 @@ A complete run command would look like this:
 docker run -d \
  --name=jellyfin \
  --gpus all \
+ -e NVIDIA_DRIVER_CAPABILITIES=all \
```
crobibero, line 237: Same

general/administration/hardware-acceleration.md
```diff
@@ -241,18 +245,21 @@ docker run -d \
...
     image: jellyfin/jellyfin
-  # ... your Jellyfin config
-  deploy:
-    resources:
-      reservations:
-        devices:
-        - capabilities: [gpu]
+    # ... your Jellyfin config
+    environment:
+      - NVIDIA_DRIVER_CAPABILITIES=all
```
crobibero, line 256: Same

lifehackerhansol: I find that weird. I could not get it working at all on my local docker-compose no matter what I do.

Should that be reported to the jellyfin repo as a bug?

Shadowghost:
We're currently already adding one of the envs per default to the container. I guess it makes more sense to add both and not put it into the docs at all.

Related PR: [jellyfin-metapackages/#22](https://github.com/jellyfin/jellyfin-metapackages/pull/22)

lifehackerhansol:
I will drop the commit, then. I suppose I'll just have to use that for personal use until it is merged.

general/administration/hardware-acceleration.md
lifehackerhansol: Dropped in favour of <https://github.com/jellyfin/jellyfin-metapackages/pull/22>

crobibero (approved):

PR #673, opened 2022-06-08
Update movies.md

siburec:
Explain that it is possible to add external subtitles to a movie.


Shadowghost:
That's already covered by the `external files` page

siburec:
> That's already covered by the `external files` page

Sorry, you are right. However, it could be useful to add a reference to that section.

seaniedan:
Please provide a link on this page, it's fiddicult to find information about subtutles. Please add where to find the list of language codes. Is it 
https://www.loc.gov/standards/iso639-2/php/code_list.php ?

PR #683, opened 2022-06-30
Added some clients

DevilsDesigns:
Android Apps
  - Jellyflut
Cross-Platform Clients
  - Sonixd
Web
  - Jellyfin Music Client

### DIFF
--- a/general/clients/index.md
+++ b/general/clients/index.md
@@ -29,6 +29,17 @@ Older browsers may be supported as a result of the needs of specific web-based c
 
 ## Android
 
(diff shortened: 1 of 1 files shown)


mcarlton00 (changes requested):
I don't think we should be linking to specific release tags here.  They're bound to get out of date before too long.  imo, just link to the github release page, not a specific one.

DevilsDesigns:
> 

their not specific they go to the release page with every release including the most current stable

PR #684, opened 2022-06-30
added theporndb plugin and manifest

DevilsDesigns:
added the porn db plugin github
and theporndb manifest


cvium:
We are not interested in having that kind of content associated with the project.

PR #687, opened 2022-07-10
Disable publish workflow on forks

nielsvanvelzen:
The workflow fails on forks (missing the token) and doesn't make sense anyway.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -14,6 +14,7 @@ on:
 jobs:
   build_publish:
     uses: ./.github/workflows/_meta.yaml
+    if: ${{ github.repository == 'jellyfin/jellyfin-docs' }}
     with:
       publish: true
     secrets:
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -14,7 +14,7 @@ on:
 jobs:
   build_publish:
     uses: ./.github/workflows/_meta.yaml
-    if: ${{ github.repository == 'jellyfin/jellyfin-docs' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -14,6 +14,7 @@ on:
 jobs:
   build_publish:
     uses: ./.github/workflows/_meta.yaml
+    if: ${{ github.repository == 'jellyfin/jellyfin-docs' }}
```
h1dden-da3m0n, line 17:
```suggestion
    if: ${{ contains(github.repository_owner, 'jellyfin') }}
```

Just as an alternative option that we could use across all applicable workflow in the org :thinking:

h1dden-da3m0n (approved):
LGTM

The suggestion is just that a suggestion, if we choose to use that or the current version will have the same result (unless the user who forks happens to have `jellyfin` within their GH user name)

.github/workflows/publish.yaml
nielsvanvelzen: This might be easier when we copy-paste workflows across repositories but it's also less strict (like you said it might false-positive on someone with jellyfin in their name, although unlikely).

I don't particularly have a preference for either of the solutions.

h1dden-da3m0n: Since it allows for easier copy-paste (and thereby less copy-pasta-errors) I have been using this in the new workflows of the CI overhaul.
So I tend to favour this less strict approach, as the likelihood of a user forking with `jellyfin` within their name is a risk I am willing to take.

Shadowghost (approved):

PR #690, opened 2022-07-14
Add nginx HTTPS example config for subpath method

le-roux:
The existing nginx config for the subpath method doesn't include the necessary elements to use HTTPS (contrary to the subdomain one).
In addition to the new server entry, the `proxy_pass` value is updated to remove the `/jellyfin/` suffix which creates an infinite redirect loop (as explained in https://www.reddit.com/r/jellyfin/comments/duhwjd/nginx_infinite_redirect_loop_after_upgrading_to/)


Shadowghost (approved):

Shadowghost:
The linter found some issues but that aside this should be fine. Didn't test it though.

le-roux:
This config has been tested OK with :
* nginx version: nginx/1.18.0 (Ubuntu)
* Jellyfin.Server 10.8.1.0

PR #694, opened 2022-08-06
Fix missing --volume in Podman install.

fisherthewol:
Commit ffc7f5317058348096d31606b4f0a304614520f9 removed line 171 from installing.md, meaning the default podman invocation fails to mount the media folder.

The commit gives no justification for this, so I assume it was an error, and propose to reintroduce it.

### DIFF
--- a/general/administration/installing.md
+++ b/general/administration/installing.md
@@ -67,8 +67,8 @@ The basic steps to create and run a Jellyfin container using Docker are as follo
     --name jellyfin \
     --user uid:gid \
     --net=host \
-    --volume /path/to/config:/config \
-    --volume /path/to/cache:/cache \
+    --volume /path/to/config:/config \ # Alternatively --volume jellyfin-config:/config
+    --volume /path/to/cache:/cache \ # Alternatively --volume jellyfin-cache:/cache
     --mount type=bind,source=/path/to/media,target=/media \
     --restart=unless-stopped \
     jellyfin/jellyfin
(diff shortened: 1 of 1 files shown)


Shadowghost:
I intentionally removed that volume mount since the only thing it would do is to create an empty named volume.
And a named volume for your media isn't what you want - you'd want a mount of a local path.

fisherthewol:
Ah, right understood.
However that change does mean that https://github.com/jellyfin/jellyfin-docs/pull/694/files#diff-9ff5cd9839e55bbf55a46c65dfe0fa82eb3051bb7696789466cbbb05b5052525L194 and https://github.com/jellyfin/jellyfin-docs/pull/694/files#diff-9ff5cd9839e55bbf55a46c65dfe0fa82eb3051bb7696789466cbbb05b5052525L196 refer to the removed line; especially for the latter line, it no longer has any context.

As well, this means the guide does not inform the user how to point Jellyfin at their media library, at least for podman - a user could infer it from the docker documentation, or have sufficient knowledge anyway, but I believe it would be worth adding it to the section directly.

I shall familiarise myself more with podman and alter this PR accordingly.

fisherthewol:
I believe the podman documentation is now accurate. I also re-did the grammar around the docker documentation, but it's worth checking how the comments inside codeblocks render, and possibly writing them outside.

Shadowghost (approved):

PR #709, opened 2022-09-15
Patch 1

psiberfunk:
Added detail about the "sledgehammer way" to reliably enable hardware acceleration for Docker on Synology.  If I figure out a better non-sledgehammer way that's more elegant, i'll update the docs.


psiberfunk:
Sorry for so many commits... had to keep correcting for the linter and then make better pictures until I was happy.

PR #710, opened 2022-09-21
Keep HTTPS request headers

BoBeR182:
https://github.com/9p4/jellyfin-plugin-sso/issues/79 fixes the above bug


Shadowghost:
Isn't this technically wrong though?
Because this headers should only be set when the connection actually is a TLS connection, but when connecting to port 80 it obviously isn't.

BoBeR182:
The rest of the file expects Let's Encrypt enabled, I see no reason why this should not be added.

Shadowghost:
Yeah you're correct, I misread it to be in th http block.

Seems like it's not possible to dynamically assign those variables like it's in nginx.

Shadowghost (approved):

PR #718, opened 2022-10-18
Document: Published Server URIs

bill-mcgonigle:
Published Server URIs doesn't seem to be documented yet.  I looked at the code and tried a test config on my networks, which seems to work at least with Android clients (i.e. different URI's reported in autodiscovery on each network).  Please improve my text as needed.


joshuaboniface (approved):

PrplHaz4:
@bill-mcgonigle Maybe link to PublishedServerURL config option or the like should go here too, so at least one of them as all ways to set a alternative published Url Value?  or vice-versa?

https://github.com/jellyfin/jellyfin-docs/blob/d4e6770ba857a4ba96be88ac8a32cbdc2115c423/general/administration/configuration.md#L99

detail via code search: https://github.com/jellyfin/jellyfin/search?q=published&type=code

PR #724, opened 2023-04-15
Update hardware-acceleration.md (Fixed typo)

ClimberJ:
Fixed typo


thornbill:
The current source of our documentation is here: https://github.com/jellyfin/jellyfin.org

This repository was moved to the archive organization because it should no longer be used.

PR #725, opened 2023-04-19
Update apache.md

mursen:
A few additions including some hints after my various failures and eventual success configuring Jellyfin tonight!


nielsvanvelzen:
> The current source of our documentation is here: https://github.com/jellyfin/jellyfin.org
> 
> This repository was moved to the archive organization because it should no longer be used.

